PR #34, opened 2016-03-30
Support leftovers, add decodeMany function

### LINKED ISSUE #33: Support leftovers in decode
If multiple messages are sent at once (e.g. when using TCP as the transport), it is currently impossible to decode all messages from a single lazy `ByteString`. Adding a variant of `decode` which returned leftovers would solve this problem. Another option would be to provide a `decodeMany` function.

I'm happy to send a PR if this sounds generally useful.

paf31:
Resolves #33 

- Add `runSGetWithLeftovers` to return leftovers after decoding using `SGet`
- Add `decodeMany` to parse many length-encoded DNS records (e.g. from TCP traffic)

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -2,9 +2,11 @@
 
 module Network.DNS.Decode (
     decode
+  , decodeMany
   , receive
   ) where
 
+import Control.Applicative (many)
 import Control.Monad (replicateM)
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
The main part is good. But please separate this pull request to `decodeMany` and other stuff such as `stack.yaml` and give me two pull requests.

paf31:
Will do, thanks!

kazu-yamamoto:
Merged. Thanks!

PR #36, opened 2016-06-16
Use Safe.toEnumMay to handle unsupported opcodes gracefully

paf31:
`toEnum` will fail with an error call if the value is not in bounds. For some DNS messages, this causes the parser to fail.


kazu-yamamoto:
LGTM. Let's merge.

paf31:
Thanks!

kazu-yamamoto:

PR #37, opened 2016-07-14
Rename a variable to fix preprocessor conflicts

niteria:
The preprocessor I use defines `unix=1` and I believe there's no harm in renaming it.


kazu-yamamoto:
I have merged and released a new version.

PR #38, opened 2016-08-18
DNS-SEC support

archaephyrryx:
Added support for TLSA record lookup and AD (authenticated data) bit to query headers

Type signatures of some internal functions (see commit notes) were changed, but this shouldn't break anything as the API is largely unmodified.

### DIFF
--- a/Network/DNS/Internal.hs
+++ b/Network/DNS/Internal.hs
@@ -137,6 +137,7 @@ data DNSFlags = DNSFlags {
   , recDesired   :: Bool
   , recAvailable :: Bool
   , rcode        :: RCODE
+  , authenData   :: Bool
   } deriving (Eq, Show)
 
 ----------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


Network/DNS/Internal.hs
```diff
@@ -216,6 +222,7 @@ instance Show RData where
   show (RD_SRV pri wei prt dom) = show pri ++ " " ++ show wei ++ " " ++ show prt ++ BS.unpack dom
   show (RD_OPT od) = show od
   show (RD_OTH is) = show is
+  show (RD_TLSA use sel mtype dgst) = show use ++ " " ++ show sel ++ " " ++ show mtype ++ " " ++ (BS.unpack $ L.toStrict . L.toLazyByteString . L.byteStringHex $ dgst)
```
archaephyrryx: Display TLSA hash data as hex string rather than raw bytes

kazu-yamamoto:
I have merged this. Thanks.

kazu-yamamoto:
We need to fix a issue of blaze-builder vs bytestring-builder. Logically, only bytestring-builder is necessary.

kazu-yamamoto:

PR #40, opened 2016-09-24
Update RFC numbers for DNAME and TLSA and add some more DNSSEC-related RRTYPEs

vdukhovni:


kazu-yamamoto:
After Haskell Symposium, my life is getting normal. Sorry for the delay.

This PR LGTM. Let's merge. Thank you for your contribution.

PR #41, opened 2016-09-26
Make it possible to sort more DNS data types

vdukhovni:
It is sometimes useful to sort RRs in a predictable order.  It would also be useful to be able to sort RData objects in their canonical wire form, because signatures on DNSSEC records are over such sorted representations of RRsets, but that is for another time...


kazu-yamamoto:
No problem. Merged.

PR #43, opened 2016-10-09
Length checks

mutjida:
We would like this change made to stop an exception from occurring-- specifically `toIPv4` from iproute calls `error` when not given exactly 4 `Int`s, similarly for `toIPv6b`.


kazu-yamamoto:
Thanks. Merged and released.

PR #45, opened 2016-10-17
Better handling of encoding and decoding the "root" domain "."

vdukhovni:
Without this, records like:

    arlott.org.uk. IN MX 0 .

fail to display correctly (the final "." is missing), and conversely
lookups of the various (e.g. NS) records of the root domain "." fail.


kazu-yamamoto:
Merged and released. Thank you for your contribution!

PR #46, opened 2016-11-05
Implemented TCP fallback after a truncated UDP response.

vdukhovni:
For example, the TLSA records for ``_25._tcp.mail.socialmiwu.de`` don't fit in a 512-byte UDP response (dig reports 596 bytes, and fails over to TCP, when EDNS0 is disabled).  The updated code is likewise able to retrieve the answer over TCP.

Please check that the implementation style matches how you want to code to be structured.  This PR strays into new territory for me, and it is less obvious what the preferred method is for reading the length bytes from a TCP socket, followed by that many bytes of response.

I should also note, that it is not clear how ParseErrors should be reported.  Should these be turned into (new?) DNS errors, or remain exceptions for the caller to `catch` (unlike DNS errors that are returned via ``Either``).  Mind you, this question is certainly not new with the addition of TCP fallback.  Malformed responses over UDP also raise the same issue.

### DIFF
--- a/Network/DNS/Resolver.hs
+++ b/Network/DNS/Resolver.hs
@@ -28,8 +28,11 @@ import Network.DNS.Decode
 import Network.DNS.Encode
 import Network.DNS.Internal
 import qualified Data.ByteString.Char8 as BS
-import Network.Socket (HostName, Socket, SocketType(Datagram), close, socket, connect)
-import Network.Socket (AddrInfoFlag(..), AddrInfo(..), SockAddr(..), PortNumber(..), defaultHints, getAddrInfo)
+import Network.Socket (HostName, Socket, SocketType(Stream, Datagram))
+import Network.Socket (AddrInfoFlag(..), AddrInfo(..), SockAddr(..))
+import Network.Socket (Family(AF_INET, AF_INET6), PortNumber(..))
+import Network.Socket (close, socket, connect, getPeerName, getAddrInfo)
+import Network.Socket (defaultHints, defaultProtocol)
 import Prelude hiding (lookup)
 import System.Random (getStdRandom, randomR)
 import System.Timeout (timeout)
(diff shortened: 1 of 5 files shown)


kazu-yamamoto:
Awesome!
I have merged this PR.

Network/DNS/Resolver.hs
```diff
@@ -342,15 +369,37 @@ lookupRawInternal rcv ad rlv dom typ = do
               Nothing  -> loop query checkSeqno (cnt + 1) False
               Just res -> do
                   let valid = checkSeqno res
-                  if valid then
-                      return $ Right res
-                    else
-                      loop query checkSeqno (cnt + 1) False
+                  case valid of
+                      False  -> loop query checkSeqno (cnt + 1) False
+                      True | False <- trunCation $ flags $ header res
```
kazu-yamamoto: This is not readable for me.
I guess that `trunCation $ flags $ header res` should be bound a variable in the `let` above and use `not var`.

kazu-yamamoto:
I wrote one comment. Please consider how to improve readability.

Network/DNS/Resolver.hs
vdukhovni: The code in question is using Haskell 2010 *pattern guards*:

https://wiki.haskell.org/Pattern_guard

Do you dislike pattern guards? I did not want to parse the headers of a message with the wrong sequence number.

Speaking of sequence timers, I think there's a problem with UDP port re-use in the library:

* Suppose a client is using a single Resolver object to make many consecutive DNS queries.
* Further suppose that one of the queries times out
* Suppose also that the server did not actually lose it, and responds shortly after.

We now have a situation where the client and server are of sync!  The server's answer will now be seen by the client in response to a different query.  The re-use of the same sequence number for query retries does not solve the problem.  After the client finally gets the expected answer there are still more answers pending than it has read, and those answers will arrive in response to subsequent queries.

Therefore, either each query needs to always create a new socket that's used only for retries of that query and then closed, or else the socket needs to be closed after any query that encounters a timeout and needs to retransmit.  The always use a new socket strategy seems simpler.

I would therefore propose that "withResolver" largely become a noop that squirrels away the seed, but does not preemptively open the socket, instead a new socket is always opened in `lookupRawInternal` and used only for a single retry loop.  The cleanup `bracket` for the socket would now happen in `lookupRawInternal`.

Thoughts?  I can implement this if you don't have a different solution in mind?

kazu-yamamoto: I did not know PatternGuards is now standard. Actually, I checked the language extensions in this file but did not found PatternGuards, so I could not understand this code.
Anyway, thank you for your explanation. I will release a new version AS IS from now on.

PR #47, opened 2016-11-08
TCP retry code cleanup

vdukhovni:
Though the pattern guard was legitimate, it was unnecessary, so I simplified it anyway.

While I was there, I took the opportunity to split off local helper functions as top-level definitions, and add comments.  This work adds no functionality and fixes no bugs, so does not merit a new release, but may be useful to stage for the next convenient opportunity.

Feedback welcome, this is supposed to make the code more clear.
Some of the comments ask questions about error handling, feedback on that especially welcome.

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -256,9 +256,9 @@ decodeRData TLSA len = RD_TLSA <$> decodeUsage
                                <*> decodeMType
                                <*> decodeADF
   where
-    decodeUsage    = getInt8
-    decodeSelector = getInt8
-    decodeMType    = getInt8
+    decodeUsage    = get8
+    decodeSelector = get8
+    decodeMType    = get8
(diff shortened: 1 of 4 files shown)


vdukhovni:
By the way, travis may need some attention... The "example.com" `A` lookup managed to complete despite a time limit of 1us, this is not too surprising, since the granularity of timeouts may not be quite that small, and answer may have arrived before the smallest time quantum expired.

Not sure what's the deal with the dependency checks in one the other build failures, but they don't appear to be related to this PR in any way.

kazu-yamamoto:
Merged. Thanks.
I don't mind the compatibility of TLSA RR.

kazu-yamamoto:
I know that test suites of the dns library are unstable.
You don't have to take care of it.

kazu-yamamoto:

PR #51, opened 2017-04-26
Added RD_DS support

vdukhovni:
This adds support for decoding, encoding and displaying DS RRs.

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -260,6 +260,17 @@ decodeRData TLSA len = RD_TLSA <$> decodeUsage
     decodeSelector = get8
     decodeMType    = get8
     decodeADF      = getNByteString (len - 3)
+--
+decodeRData DS len = RD_DS <$> decodeTag
+                           <*> decodeAlg
+                           <*> decodeDtyp
+                           <*> decodeDval
+  where
+    decodeTag  = get16
+    decodeAlg  = get8
+    decodeDtyp = get8
+    decodeDval = getNByteString (len - 4)
+--
 decodeRData _  len = RD_OTH <$> getNByteString len
 
 decodeOData :: OPTTYPE -> Int -> SGet OData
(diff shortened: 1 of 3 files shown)


kazu-yamamoto:
I'm not sure that the CI failure is relating to your code. Any thoughts?

vdukhovni:
It sure looks unrelated.
```
cabal: Error: some packages failed to install:
safe-0.3.14 failed during the configure step. The exception was:
user error (The package 'safe' requires Cabal library version -any && >=1.18
but no suitable version is installed.)
```
I guess we're now using a newer default version of `safe` that needs a newer GHC than 7.6.3?  Perhaps that configuration needs a lower maximum version of `safe`?  I'd expect the master branch to fail with 7.6.3 without my PR...

vdukhovni:
Looks like `safe-0.3.12` changed to depend on `base (>=4.5 && <5)` while `0.3.11` had just `base (<5)`.
The changelog says:
```
 0.3.14
   #20, fix for GHC 7.10.1
0.3.13
   #20, require GHC 7.4 or above
0.3.12
  #19, add Safe.Partial exposing a Partial typeclass
  #19, add support for GHC call stacks
```
However, it seems that 7.4 may not be enough, or there's some other constraint that failed to be recorded.

vdukhovni:
Could you look into it further?
https://github.com/ndmitchell/safe/issues/20

kazu-yamamoto:
We don't have to keep GHC 7.6.
Also, this PR looks good to me.
So, merged. Thanks.
I would release a new version if you wish.

vdukhovni:
Thanks.  No rush on a new release, I can keep using a private build for a while, perhaps there'll be other minor changes before you need to cut a new release.  Likely not too many folks other than me are getting their hands dirty with DNSSEC in Haskell.  There are more RRtypes we could add decoders for, I've not yet been inspired to start adding `RD_sometype` definitions for RRtypes I don't personally use.

kazu-yamamoto:
Please ping me when you would like to have a new release.

Issue #27, opened 2015-04-21
The identifier of DNSHeader can only hold 16 bits integers.


redneb:
I was debugging a misbehaving program of mine today and I found out that the bug was caused by writing an integer greater than 65535 to the `identifier` of a `DNSHeader`. Currently that field has type `Int` but a DNS message has room for only 16 bits to written there. The problem is that `encode` (from `Network.DNS.Encode`) silently performs a lossy conversion discarding the excess bits. This can lead to various unexpected bugs.

Maybe `encode` could be modified so that it raises an exception if a value larger than 65535 is provided. Or the type of that field could be changed to `Word16`. I would prefer the latter, but unfortunately that would be a breaking change.

kazu-yamamoto:
I also like `Word16` and think its effect is small because it is a low API. Let's bump up the version to 1.5 to express this breaking change. Would you give me a pull request?

redneb:
> Would you give me a pull request?

Will do.

PR #52, opened 2017-06-17
Change DNSHeader identifier field to Word16

erikd:
The `identifier` field was `Int`, but the specs say its `Word16` so
its better to use `Word16`.

Closes: https://github.com/kazu-yamamoto/dns/issues/27

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -140,7 +140,7 @@ decodeHeader = do
              ,arCount
              )
   where
-    decodeIdentifier = getInt16
+    decodeIdentifier = get16
     decodeQdCount = getInt16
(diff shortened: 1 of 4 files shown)


erikd:
This is an API change and would require a major version bump.

erikd:
I'm currently writing some QuickCheck round-trip tests which may suggest some changes to the data types. I therefore suggest you don't release a new version until that is sorted out :).

kazu-yamamoto:
@erikd May I merge this PR anyway?

erikd:
I should probably close this PR and submit the commit here as part of PR #53 . Just to keep all the API breaking changes in a single PR.

kazu-yamamoto:
OK. Thanks.

PR #53, opened 2017-06-17
Topic/round trip tests

erikd:
Add round trip encode/decode tests.

~~~On top of~~~ Incorporates https://github.com/kazu-yamamoto/dns/pull/52

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -2,6 +2,10 @@
 
 module Network.DNS.Decode (
     decode
+  , decodeDomain
+  , decodeDNSFlags
+  , decodeDNSHeader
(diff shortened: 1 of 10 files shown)


erikd:
Sent a PR too soon. This is not ready yet.

erikd:
@kazu-yamamoto I think this is ready now. Everything can can now round-trip except for the `OptRecord` constructor of `ResourceRecord`. Reading [rfc6891]https://tools.ietf.org/html/rfc6891#section-6.1.2) I think the current definition of `OptRecord` is not correct and should actually be something like:
```
    OptRecord Word16 Bool Word8 [RData]
```

kazu-yamamoto:
The code got more elegant. Thanks!

PR #57, opened 2017-07-01
Fix the StateBinary.get32 parser

erikd:
Found when round-tripping packets captured with Wireshark.

### DIFF
--- a/Network/DNS/StateBinary.hs
+++ b/Network/DNS/StateBinary.hs
@@ -125,7 +125,7 @@ get16 = ST.lift getWord16be <* addPosition 2
     getWord16be = do
         a <- word8'
         b <- word8'
-        return $ a * 256 + b
+        return $ a * 0x100 + b
 
 get32 :: SGet Word32
 get32 = ST.lift getWord32be <* addPosition 4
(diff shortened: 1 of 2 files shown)


Network/DNS/StateBinary.hs
```diff
@@ -125,7 +125,7 @@ get16 = ST.lift getWord16be <* addPosition 2
     getWord16be = do
         a <- word8'
         b <- word8'
-        return $ a * 256 + b
+        return $ a * 0x100 + b
```
kazu-yamamoto: Is this a just style change?

Network/DNS/StateBinary.hs
```diff
@@ -136,7 +136,7 @@ get32 = ST.lift getWord32be <* addPosition 4
         b <- word8'
         c <- word8'
         d <- word8'
-        return $ a * 1677721 + b * 65536 + c * 256 + d
+        return $ a * 0x1000000 + b * 0x10000 + c * 0x100 + d
```
kazu-yamamoto: Does this mean that I made a mistake on copying `16777216`?

Network/DNS/StateBinary.hs
erikd: Yes, most coders know that `256` is `0x100 but few coders can tell the decimal value of `0x1000000`. Changing this one so its in the same form as the others below.

Network/DNS/StateBinary.hs
erikd: I think you know the answer to that :).

kazu-yamamoto: Thank you for finding this. :-)

kazu-yamamoto:
Merged.
Thank you for your contributions!

Network/DNS/StateBinary.hs
erikd: Thank you for providing this library. *Soooo* much easier to extend this than starting from scratch.

PR #59, opened 2017-07-03
Switch to strict ByteString

erikd:
DNS packets are small and there is little to no advantage of using
lazy ByteStrings.


erikd:
Ping?

kazu-yamamoto:
Sorry for the delay. I did not notice this PR. Merged.

PR #60, opened 2017-07-04
Update gitignore

keizo042:
hello.
I use codex that ctag file generator for cabal project dependencies.
I'd like to let git ignore tags file and object file that relate to examples.


kazu-yamamoto:
Merged. Thank you!

PR #61, opened 2017-07-22
Remove bytestring-builder dependency.

QuLogic:
It was removed in b00335f8b0a6e1c287f87450190fda266968ae58.

Or at the very least, according to its docs, it's merged into GHC 7.8, and since the Travis build here dropped 7.6, I'd guess it shouldn't be necessary.


kazu-yamamoto:
Merged. Thank you for telling me this.

PR #69, opened 2017-08-23
Revert ResourceRecord API maintaining compatibility with prior releases.

vdukhovni:
This should address issue #68

Issue #68, opened 2017-08-23
Disappearance of `rrname`, `rrtype`, `rrttl` and `rdata` accessors of ResourceRecord


vdukhovni:
Some recent commits have removed the "record" form of ResourceRecord, leaving only a multi-slot constructor.  Existing code that used the accessors must now change to define its own accessors.  Only getRdata has been provided.

What motivates these changes?  They seem ill-advised.  In particular they fundamentally break the previous API and all existing applications using Network.DNS would have to change.  I think that these changes should also be reverted...

kazu-yamamoto:
See #63.

vdukhovni:
Well, given the existing API, and the rather meager gains in safety by avoiding partial functions (since most applications will never handle DNSMessage objects and will just deal with answers and perhaps authority and additional records where OPT records don't appear), my contention is that #63 is misguided, and does not adequately justify breaking API.

A related major problem is the disappearance of the custom Show instances for RData, these need to come back.

vdukhovni:
Pull request #70 addresses this and other issues, and passes all tests (a couple of tests needed to be fixed, and others have also opened smaller pull requests for a subset of the test failures).  There is no visible breakage from having separate accessors for the different branches of the ResourceRecord sum type.

kazu-yamamoto:
I have merged #70.
But `ResourceRecord` is not a sum type anymore.
Accessors came back, so minimum backward compatibility recovered.
@vdukhovni I would like to hear your opinion.

vdukhovni:
Thanks, one quick observation... I notice you did not merge the change that changed the 1 ms timeout to 0ms to test a forced timeout.  Unfortunately, on my system with a local resolver, a 1 ms timeout does not reliably fail to perform the lookup, while 0ms does work.  Please consider adopting that change

```
 --   >>> let hostname = Data.ByteString.Char8.pack "www.example.com"
---   >>> let badrc = defaultResolvConf { resolvTimeout = 0, resolvRetry = 1 }
```

vdukhovni:
Otherwise, my initial impression (not a thorough code review) is that the changes look fine.  Thanks.

vdukhovni:
FWIW, my code compiles without warnings and seems to run correctly against the master DNS branch.  Below is how it reports the DS, DNSKEY and SOA records for "mil." (chosen in part for its "interesting" SOA mailbox name).

```
[...]

PR #70, opened 2017-08-24
More complete SOA implementation and subsume 67 and 69

vdukhovni:
This pull requests implements a more complete "show" for SOA records by properly displaying the "mrname" as an email address, and correctly displays cases where the first label contains an internal "." (first.last@domain).  It also subsumes PR #67 and PR #69 which are pre-requisites.

TODO: Even for (normail not mailbox) domains the encoder and decoder should be able to handle labels with embeded `.` characters, and when decoding escape these with a backslash per the presentation form specified in https://tools.ietf.org/html/rfc4343#section-2.1, when encoding the backslashes would need to suppress label breaks.

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -3,6 +3,7 @@
 module Network.DNS.Decode (
     decode
   , decodeDomain
+  , decodeMailbox
   , decodeDNSFlags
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
Merged. I would close this issue but continue the discussion in #68.

PR #72, opened 2017-08-30
Fix sendAll on Windows

adinapoli-iohk:
This PR fixes some lookup problems Windows users were facing when
using the library. The culprit (as confirmed by some tests) was in
the implementation of `sendAll`, that on Windows was defined as part of
the code of this library, but this is not necessary anymore unless an
old version of GHC (< 7) is being used.

For modern environments, `Network.Socket.ByteString.sendAll` is all we
need.

### DIFF
--- a/Network/DNS/Lookup.hs
+++ b/Network/DNS/Lookup.hs
@@ -183,7 +183,7 @@ lookupMX rlv dom = do
 --   >>> rs <- makeResolvSeed defaultResolvConf
 --   >>> ips <- withResolver rs $ \resolver -> lookupAviaMX resolver hostname
 --   >>> fmap sort ips
---   Right [133.138.10.34,203.178.136.49]
+--   Right [133.138.10.39,203.178.136.30]
 --
 --   Since there is more than one result, it is necessary to sort the
 --   list in order to check for equality.
--- a/Network/DNS/Resolver.hs
+++ b/Network/DNS/Resolver.hs
@@ -19,6 +19,16 @@ module Network.DNS.Resolver (
   , fromDNSFormat
   ) where
 
+#if !defined(mingw32_HOST_OS)
+#define POSIX
+#else
+#define WIN
+#endif
+
+#if __GLASGOW_HASKELL__ < 709
+#define GHC6
+#endif
+
 import Control.Exception (bracket)
 import Data.Char (isSpace)
 import Data.List (isPrefixOf)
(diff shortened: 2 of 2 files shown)


adinapoli-iohk:
cc @kazu-yamamoto @dcoutts

kazu-yamamoto:
The dns library supports only three major versions.
Do you have any strong reason to support GHC 6.x?

adinapoli-iohk:
@kazu-yamamoto hey!

Sorry, I think the mistake was fine for naming the `CPP` constant poorly: when I saw the expression:

```
#if __GLASGOW_HASKELL__ < 709
```

I must have thought we were implying some oldish version of GHC like 6.x, thus my ill-named `GHC6` constant. What about I rename that simply `GHC709`?

> Do you have any strong reason to support GHC 6.x?

I do not! In reality I was just taken in by the conditional `build-depends` in the cabal manifest:

```
  if impl(ghc >= 7)
    Build-Depends:      base >= 4 && < 5
                      , attoparsec
                      , binary
                      , bytestring
                      , conduit >= 1.1
                      , conduit-extra >= 1.1
                      , containers
                      , iproute >= 1.3.2
                      , mtl
                      , network >= 2.3
                      , random
                      , resourcet
                      , safe == 0.3.*
```

As far as I'm concerned, I'm perfectly happy to drop those and simply remove the custom `sendAll` once for all 😉

adinapoli-iohk:
@kazu-yamamoto To keep this "hot": Which are the changes you would like to see applied to this PR to make it mergeable into `master`? Thank you! 😉

kazu-yamamoto:
> What about I rename that simply GHC709?

Good.

Would you clean up this PR, rebase onto master and do `push -f`?
I would like to merge it.

P.S. I setup AppVoyer which provides GHC 7.10.3.
Do you know how to use multiple GHC versions on AppVoyer?

kazu-yamamoto:
Wow. Your patch fixes the test failure on AppVoyer. Great.
So, I will treat this patch by myself.

adinapoli-iohk:
> Wow. Your patch fixes the test failure on AppVoyer. Great.

Aye, and we are successfully using it on production for DNS lookups on Windows 😉 

> Do you know how to use multiple GHC versions on AppVoyer?

I don't know if `stack` is your cup of tea, but using it on Appveyor for multiple versions of GHC is quite easy! For example: https://github.com/lpeterse/haskell-socket/blob/master/appveyor.yml

kazu-yamamoto:
Merged. And multiple GHCs are supported on AppVeyor.

adinapoli-iohk:
Thank you very much, @kazu-yamamoto !

kazu-yamamoto:
The next release of `master` is a major. So, we should be carefully.

But if you want to merge your fix into to the `stable-2.0` branch, I can release a 2.0.13 very quickly.
What do you think?

adinapoli-iohk:
> But if you want to merge your fix into to the stable-2.0 branch, I can release a 2.0.13 very quickly. What do you think?

Hey! Just to be sure to be on the same page, can I summarise what you are saying? You are telling me that releasing the current `master` is a major release, so it shouldn't be rushed, but that we can still release `2.0.13` relative quickly if we drop the default DNS thing and we include only:

* The `RetryLimitExceeded` patch;
* The `sendAll` fix patch;

Am I right?

kazu-yamamoto:
Yes!

adinapoli-iohk:
Works for me!

PR #77, opened 2017-09-07
Permit max-length domains with trailing dot

vdukhovni:
Eventually, when we implement "de-escaping" domains on input (and escaping on output), we'll need to convert all the labels to wire form in order to determine whether a domain is legal or not.  Then we can just check whether `any ((> 63) . length) labels` or `sum (map length labels) + length labels + 1 > 255`.

Conversion of labels to wire form can reduce their length: `"foo\.bar" -> "foo.bar"`

For now, since we never de-escape, just checking the Domain length suffices, but we need to allow one more byte for domains that already have a trailing dot (or really one less for those that don't, but that's the current code).


kazu-yamamoto:

PR #85, opened 2017-09-20
Catch and annotate IO Exceptions in performLookup

vdukhovni:
When IO errors happen, it is useful to know where.  We annotate any
I/O exceptions with the protocol, address and port


kazu-yamamoto:
Merged.

PR #87, opened 2017-09-25
Use `drop 1` instead of `tail`

themattchan:
This fixes the TXT  parser so that it does not fail on empty input.

I have an additional question regarding the semantics of `StateBinary.getNByteString` -- is the position in `SGet` meant to track the logical position or the number of bytes actually consumed? For example, `getNByteString 10` will consume only 5 bytes if fed a 5 byte input but the resulting offset inside the `SGet` would be 10, which seems incorrect to me.

### DIFF
--- a/Network/DNS/Decode.hs
+++ b/Network/DNS/Decode.hs
@@ -208,7 +208,7 @@ getRData CNAME _ = RD_CNAME <$> getDomain
 getRData DNAME _ = RD_DNAME <$> getDomain
 getRData TXT len = (RD_TXT . ignoreLength) <$> getNByteString len
   where
-    ignoreLength = BS.tail
+    ignoreLength = BS.drop 1
 getRData A len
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
> I have an additional question regarding the semantics of StateBinary.getNByteString -- is the position in SGet meant to track the logical position or the number of bytes actually consumed? For example, getNByteString 10 will consume only 5 bytes if fed a 5 byte input but the resulting offset inside the SGet would be 10, which seems incorrect to me.

I believe that it is the number of bytes actually consumed.
Would you add a commit to fix this issue to this PR by yourself?

themattchan:
Thanks for the clarification -- I'll open a separate commit to address that, since it is a separate concern to this bugfix.

kazu-yamamoto:
Rebased and merged.
I wait for another PR from you.

PR #89, opened 2017-10-04
Use a fresh UDP socket for each query.

vdukhovni:
This avoids the possibility of getting out of sync with the namerver
after (perhaps multiple) timeouts.  It also avoids idle file
descriptors, one per nameserver sitting around between queries.
Also we get some more protection from Kaminsky-style attacks, though
we still need a better RNG for query ids, getting a random port
number is left up to the OS, this is too expensive to do in user-land.

### DIFF
--- a/Network/DNS/Resolver.hs
+++ b/Network/DNS/Resolver.hs
@@ -140,7 +140,7 @@ data ResolvSeed = ResolvSeed {
 --   When implementing a DNS cache, this MUST NOT be re-used.
 data Resolver = Resolver {
     genId      :: IO Word16
-  , dnsSocks   :: NonEmpty Socket
+  , dnsServers :: NonEmpty AddrInfo
   , dnsTimeout :: Int
   , dnsRetry   :: Int
   , dnsBufsize :: Integer
@@ -213,44 +213,30 @@ makeAddrInfo addr mport = do
 
 
 -- | Giving a thread-safe 'Resolver' to the function of the second
---   argument. A socket for UDP is opened inside and is surely closed.
---   Multiple 'withResolver's can be used concurrently.
+--   argument.  Multiple 'withResolver's can be used concurrently.
 --   Multiple lookups must be done sequentially with a given
---   'Resolver'. If multiple 'Resolver's are necessary for
---   concurrent purpose, use 'withResolvers'.
+--   'Resolver'. If multiple 'Resolver's are needed concurrently,
+--   use 'withResolvers'.
 withResolver :: ResolvSeed -> (Resolver -> IO a) -> IO a
-withResolver seed f = bracket (initSockets seed) (mapM_ close) (f . makeResolver seed)
+withResolver seed f = f $ makeResolver seed
 
 -- | Giving thread-safe 'Resolver's to the function of the second
---   argument. Sockets for UDP are opened inside and are surely closed.
---   For each 'Resolver', multiple lookups must be done sequentially.
---   'Resolver's can be used concurrently.
+--   argument.  For each 'Resolver', multiple lookups must be done
+--   sequentially.  'Resolver's can be used concurrently.
 withResolvers :: [ResolvSeed] -> ([Resolver] -> IO a) -> IO a
-withResolvers seeds func = bracket (mapM initSockets seeds) closeSockets $ \socks -> do
-    let resolvs = zipWith makeResolver seeds socks
-    func resolvs
-  where
-    closeSockets = mapM_ close . concatMap NE.toList
+withResolvers seeds f = f $ map makeResolver seeds
 
--- | Atomically create and connect to all the nameservers listed in the given
--- `ResolvSeed`.
-initSockets :: ResolvSeed -> IO (NonEmpty Socket)
-initSockets seed = mapM initSocket addrs
-  where
-    initSocket ai = do sock <- socket (addrFamily ai) (addrSocketType ai) (addrProtocol ai)
-                       connect sock (addrAddress ai)
-                       return sock
-    addrs         = nameservers seed
-
-makeResolver :: ResolvSeed -> NonEmpty Socket -> Resolver
-makeResolver seed socks = Resolver {
+makeResolver :: ResolvSeed -> Resolver
+makeResolver seed = Resolver {
     genId = getRandom
-  , dnsSocks = socks
+  , dnsServers = nameservers seed
   , dnsTimeout = rsTimeout seed
   , dnsRetry = rsRetry seed
   , dnsBufsize = rsBufsize seed
   }
 
+-- | XXX: This is unlikely to be cryptographically strong.  We should use a
+-- generator that resists cryptanalysis.
 getRandom :: IO Word16
 getRandom = getStdRandom random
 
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
Code looks good to me.

@vshabanov Would you test the performance?

vdukhovni:
One question I am not 100% sure about is whether the current `genId` RNG is thread-safe.  It looks thread-safe to me based on its implementation via `atomicallyModifyIORef'` in `System.Random`, but various answers on stack-exchange seem to suggest otherwise.  Perhaps they are considering more general StdGen instances than the global one, which is protected by a global atomic critical section it seems to me.

It does create a global critical section, but this probably is cheap enough.  If we use `split` to get a per-thread RNG, we could avoid that particular global lock, but then we'd need a per-resolver RNG, and lose the opportunity (see below) to use a single resolver in multiple threads!

As it stands, after the patch we no longer require a per-thread resolver. With the sockets created on demand, everything is shareable.  We could now provide a method to obtain a long-term resolver handle that is not wrapped around a single action, the object in question holds nothing that needs finalization and can just be garbage-collected as appropriate.

The `withResolvers` API also ultimately can be deprecated.  But all this later, first we just decide whether socket-per-request is appreciably slower (and avoiding the cost warrants the hazards of reuse).

Network/DNS/Resolver.hs
```diff
@@ -450,14 +443,14 @@ lookupRawInternal rcv ad rlv dom typ = loop (NE.uncons (dnsSocks rlv))
...
-         -> Socket
+         -> AddrInfo
          -> Int
          -> IO (Either DNSError DNSMessage)
-tcpRetry query sock tm = do
-    peer <- getPeerName sock
-    bracket (tcpOpen peer)
+tcpRetry query ai tm = do
+    let addr = addrAddress ai
+    bracket (tcpOpen $ addr)
```
vdukhovni: I should probably drop the extraneous `$` before this is merged...

Network/DNS/Resolver.hs
```diff
@@ -397,23 +383,30 @@ lookupRawInternal ::
...
+                (close)
+                (\sock -> performLookup ai sock query checkSeqno 0 False)
       case res of
-        Left e  -> maybe (return (Left e)) (loop . NE.uncons) alternatives
+        Left e  -> maybe (return (Left e)) (loop . NE.uncons) ais
         Right v -> pure (Right v)
 
+    udpOpen ai = do
+        sock <- socket (addrFamily ai) (addrSocketType ai) (addrProtocol ai)
+        connect sock (addrAddress ai)
```
vdukhovni: There's a small chance of raising an uncaught I/O error here, if socket() fails due to file-descriptor exhaustion (system-wide or per-process limits). Same issue applied to `withResover` in the original code, we could attempt to address this.  Likewise the connect() could fail due to failure acquire a free local port, and then we also leak the socket.  Both conditions are rather uncommon.  Perhaps just a comment for now, and some appropriate error handling later. We could return "NetworkError" or create a new branch of the DNSError sum type for SystemError (these are local resource problems).

Network/DNS/Resolver.hs
vdukhovni: `tcpOpen` can fail in much the same way as `udpOpen`, though this time `connect()` failure is in caught separately in `tcpLookup`, and is generally an actual network problem.

kazu-yamamoto: Yeah.

vdukhovni:
I hope you don't mind the self-review.  These are comments about things I did not choose to improve while delaying socket creation. They are largely existing issues, that perhaps should be addressed, and are not particularly difficult.  Thoughts?

kazu-yamamoto:
> One question I am not 100% sure about is whether the current genId RNG is thread-safe. It looks thread-safe to me based on its implementation via atomicallyModifyIORef' in System.Random, but various answers on stack-exchange seem to suggest otherwise. Perhaps they are considering more general StdGen instances than the global one, which is protected by a global atomic critical section it seems to me.

Good point. We should replace the `random` package simply because it is slow.

@kosmikus Would you suggest a good random library which is both fast and thread-safe?

Network/DNS/Resolver.hs
kazu-yamamoto: > Perhaps just a comment for now, and some appropriate error handling later. 

Agree.

Network/DNS/Resolver.hs
vdukhovni: Fixed in the next commit.

Network/DNS/Resolver.hs
vdukhovni: Comments for udpOpen and tcpOpen fixed in next commit.

vdukhovni:
Shall we proceed with this PR?  It seems feedback on performance impact (if any) is not forthcoming...

vshabanov:
Sorry, have no time to test performance at the moment. But I use local BIND (and hence local sockets) so I don't think that performance would be affected much.

If you like to test performance difference I've extracted a list of domain names from my feeds fetcher log:
https://beta.bazqux.com/web/domains.txt.bz2

There are 100k+ unique domains (some are invalid) repeated many times as feeds are refetched. You could try to run old a new version over this list several times to see whether there is any major difference.

vdukhovni:
> @vshabanov There are 100k+ unique domains (some are invalid) repeated many times as feeds are refetched. You could try to run old a new version over this list several times to see whether there is any major difference.

Sadly, it is not really possible to simulate the performance impact (I don't expect any) on your application outside the application.  There are key factors like your cache hit rate, the query rate, how the application parallelises queries, how the kernel allocates sockets, ... that make it very difficult to predict whether you'll see any difference at all, and whether it'll be significant.

vdukhovni:
My take is that given that the change has benefits, and fixes  a possible break-down of synchronization between the DNS client and server, the changes should be merged, and the cost should be acceptable.

vshabanov:
> and fixes a possible break-down of synchronization between the DNS client and server

IIRC I've had this issue when updated (restarted?) BIND with working crawler -- it silently stopped resolving anything. Although I don't remember all details it looks related to socket that is never reopened.

When I've read about Kaminsky attack it looks that some servers use pool of preopened sockets. It could be an option to improve performance. Although, again, I'm not sure that performance is affected much. Anyway I'm doing and HTTP connection to download feed, then process it and update database so additional opened/closed socket is not an issue. I'm much more interested in stability and correctness not the performance.

kazu-yamamoto:
I'm now trying to run https://github.com/kazu-yamamoto/concurrent-dns-cache/blob/master/main.hs with @vshabanov's domain list to see the performance.

vdukhovni:
> @kazu-yamamoto I'm now trying to run https://github.com/kazu-yamamoto/concurrent-dns-cache/blob/master/main.hs with @vshabanov's domain list to see the performance.

I am puzzled as to how this measures anything relevant... The code in questions seems to use Google's resolvers, which:

* May rate limit queries as a DoS counter-measure.
* Are probably not sufficiently close (e.g. loopback) to make the latency of socket creation measurable...

Nor is it clear how 50 "threads" doing parallel back to back queries are representative of a configuration doing other application work, with the caching resolver at 127.0.0.1.  It would probably be more interesting to spin up an unbound resolver on the loopback interface (if you don't already have one) and use that, and still the relevance of the measurement will likely be unclear...

kazu-yamamoto:
Here are results:

Current: 1278000 domains: 1156.593191 sec
With PR: 1278000 domains: 1109.343742 sec

kazu-yamamoto:
> Nor is it clear how 50 "threads" doing parallel back to back queries are representative of a configuration doing other application work, with the caching resolver at 127.0.0.1. It would probably be more interesting to spin up an unbound resolver on the loopback interface (if you don't already have one) and use that, and still the relevance of the measurement will likely be unclear...

I understand your concern but I have no future idea to measure the performance at this moment.

kazu-yamamoto:
@vdukhovni I don't know that my experiment is meaningful but I'm now OK to merge this PR.

vdukhovni:
Thanks, much appreciated,  I hope my tone does not come across as confrontational, that was not the idea. Just trying to point out that I think that performance benchmarks are difficult to make broadly applicable. Please merge (or rebase and merge) the PR at your convenience.

kazu-yamamoto:
> Thanks, much appreciated, I hope my tone does not come across as confrontational, that was not the idea. Just trying to point out that I think that performance benchmarks are difficult to make broadly applicable.

No problem!

> Please merge (or rebase and merge) the PR at your convenience.

PR #99, opened 2017-11-06
Reworked UDP transport loop.

vdukhovni:
It is now possible to retry without EDNS0 at the same retry counter
without risking infinite loops.  Also comment on NotImpl wrt. EDNS0
retries.  The new code is IMHO somewhat cleaner.

Also handle DNS id mismatches over TCP.  The new checkResp function
may be some day useful for pipelined TCP queries with out-of-order
responses for matching the response to some outstanding request.

### DIFF
--- a/Network/DNS/Transport.hs
+++ b/Network/DNS/Transport.hs
@@ -7,16 +7,22 @@ module Network.DNS.Transport (
   ) where
 
 import Control.Exception as E
-import Data.ByteString (ByteString)
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.List.NonEmpty as NE
 import Data.Typeable
+import Data.Word (Word16)
 import Network.DNS.IO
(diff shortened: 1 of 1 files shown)


vdukhovni:
Question, why was the `NotImpl` case added for retrying EDNS0?  Do you know of recursive servers that in-fact botch EDNS0 by returning `NotImpl` and not `FormErr`?

kazu-yamamoto:
> Question, why was the NotImpl case added for retrying EDNS0? Do you know of recursive servers that in-fact botch EDNS0 by returning NotImpl and not FormErr?

This is because I found technical slides about DNS written in Japanese. One page says NotImpl or FormErr. But I'm not sure that NotImpl actually happens. So, I'm OK to remove it.

kazu-yamamoto:
Merged. NotImpl is removed.

PR #101, opened 2017-11-18
Remove minttl and cosmetic fixes

vdukhovni:
I don't think there should be a minimum TTL.  If we get a TTL less than 60s, we should either honour it or not cache the data, but we should not raise the received TTL.  My preference is to honour whatever TTL we get.


kazu-yamamoto:
Make sense to me. Thank you for your contribution!

# I have accidentally pushed the merge commit before approving this. Sorry.

PR #105, opened 2018-03-06
catching up conduit 1.3.

kazu-yamamoto:
@snoyberg May I ask you to review this PR?

### DIFF
--- a/Network/DNS/IO.hs
+++ b/Network/DNS/IO.hs
@@ -31,11 +31,7 @@ import qualified Data.ByteString.Builder as BB
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import Data.Char (ord)
-#if MIN_VERSION_conduit(1,3,0)
 import Data.Conduit (($$+), ($$+-), ConduitM, (.|), runConduit)
-#else
-import Data.Conduit (($$+), ($$+-), ConduitM, (.|), ($$))
-#endif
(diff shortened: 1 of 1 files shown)


Network/DNS/IO.hs
```diff
@@ -53,13 +57,21 @@ import Network.DNS.Types
 
 ----------------------------------------------------------------
 
+#if MIN_VERSION_conduit(1,3,0)
+sink :: ConduitT ByteString Void IO (DNSMessage, PState)
```
snoyberg: If you want something which will work on both new and old conduit, you can use:

    ConduitM ByteString Void IO (DNSMessage, PState)

Also, you can probably generalize the type signature to:

    ConduitM ByteString o IO (DNSMessage, PState)

Network/DNS/IO.hs
```diff
@@ -70,7 +82,11 @@ receiveVC sock = do
     let len = case map ord $ LBS.unpack lenbytes of
                 [hi, lo] -> 256 * hi + lo
                 _        -> 0
+#if MIN_VERSION_conduit(1,3,0)
+    fst <$> (src $$+- CB.isolate len .| sink)
```
snoyberg: The `.|` operator is available in conduit 1.2 as well (starting with 1.2.8), if you'd like to bypass the CPP here. Same applies to the implementation of `receive`.

kazu-yamamoto:
@snoyberg Big thanks to you! I was able to remove `MIN_VERSION`.

PR #111, opened 2018-04-17
Improving decoder

### LINKED ISSUE #103: Compression popping values that have not been pushed.
This snippet:

```
  rs <- makeResolvSeed defaultResolvConf {
          resolvInfo = RCHostName "90.155.94.93"
         }
  r <- withResolver rs $ \resolver -> lookupRaw resolver "www.cqx.ltd.uk" RRSIG

  print r
```

generates the attached pcap, and then fails with:

```
dugnutt: ParseError {errorContexts = [], errorMessage = "Failed reading: getDomain: 338", errorPosition = 1:449 (448)}
```

I put in some tracing around the push/pop statements and it looks like that location is not being `push`ed.

```
dugnutt www.cqx.ltd.uk RRSIG 
BENC: d pushing pos 24
BENC: d pushing pos 20
BENC: d pushing pos 16
BENC: d pushing pos 12
BENC: c pushing pos 32
BENC: c pushing pos 170
BENC: c pushing pos 308
dugnutt: ParseError {errorContexts = [], errorMessage = "Failed reading: getDomain': failed to resolve compression pointer to offset 338 at position 446", errorPosition = 1:449 (448)}
```

That server 90.155.94.93 is under my control and running bind 9.8.4.

pcap of the interaction:

www.hawaga.org.uk/tmp/dns-rrsig.pcap

### LINKED ISSUE #107: Handling RData parsing errors
After months of running daily without problems my scanner threw an exception today due to unexpected malformed data from a remote zone:
```
ParseError {errorContexts = [], errorMessage = "Failed reading: IPv6 addresses must be 16 bytes long", errorPosition = 1:81 (80)}
```
This error is from `Data.Conduit.Attoparsec`. but that library is (correctly) not a documented dependency of `Network.DNS` and I think that its internal types should not be visible to `Network.DNS` users. And yet getting garbage in the RData from some remote DNS server is not an entirely unexpected condition. Mostly one sees in `Domain` fields we don't validate very strictly, so no exceptions are raised, for example, at the moment:
```
promitterra.nl. IN MX 0 http://fb1.dootall.com.
promitterra.nl. IN MX 10 mr1.dootall.com.
;
spirit-art.nl. IN CNAME http//www.vanhoftenit.nl.
```
However, sometimes the garbage is more structural, as with unexpected RData lengths for AAAA records.

I see two possible ways to handle this more appropriately than having unexpected exceptions from the conduit parser library.  IIn either case catch parser exceptions while parsing RData from packets.  Then, either:
* If it fails to parse, create an `RD_PARSERR Int16 ByteString String` object which captures the rtype, the raw payload that could not be parsed and the error message from the parser library, or else
* If it fails to parse return "Left DNS.ParseError" for the whole DNS response

My preference is very much for the first option since one then still gets to see any valid records that are returned allong with the malformed record, and also the maformed payload and its rtype. That said, if it is much simpler to just invalidate the whole reply, given that malformed RData is comparatively rare, I could live with that for now, and the more detailed approach could be taken later...

Thoughts?  Comments?  Patches?

kazu-yamamoto:
I hope this PR fixes #103 and #107.

- The decoder now holds the original input and can parse domains even if the structure of the target record is unknown. See #103 for more information. Note that I have not implemented `RRSIG`.
- The decoder now throws `DNSError` to fix #107. I'm sorry but it appeared that the `RD_PARSERR` approach is very difficult.

@benclifford and @vdukhovni Would you review this PR?

### DIFF
--- a/Network/DNS/Decode/Internal.hs
+++ b/Network/DNS/Decode/Internal.hs
@@ -218,16 +218,20 @@ getOData opc len = UnknownOData opc <$> getNByteString len
 ----------------------------------------------------------------
 
 getDomain :: SGet Domain
-getDomain = getDomain' '.' []
+getDomain = do
+    lim <- B.length <$> getInput
+    getDomain' '.' lim []
 
 getMailbox :: SGet Mailbox
-getMailbox = getDomain' '@' []
+getMailbox = do
+    lim <- B.length <$> getInput
+    getDomain' '@' lim []
 
 -- | Get a domain name, using sep1 as the separate between the 1st and 2nd
 -- label.  Subsequent labels (and always the trailing label) are terminated
 -- with a ".".
-getDomain' :: Char -> [Int] -> SGet ByteString
-getDomain' sep1 pointerStack = do
+getDomain' :: Char -> Int -> [Int] -> SGet ByteString
+getDomain' sep1 lim pointerStack = do
     pos <- getPosition
     c <- getInt8
     let n = getValue c
@@ -237,15 +241,15 @@ getDomain' sep1 pointerStack = do
         _ | isPointer c -> do
             d <- getInt8
             let offset = n * 256 + d
+            when (offset >= lim) $ fail "pointer is too large"
             mo <- pop offset
             case mo of
                 Nothing
                   | offset `elem` pointerStack -> return "" -- pointer loop
                   | otherwise                  -> do
-                      inp <- getInput
+                      target <- B.drop offset <$> getInput
                       let newStack = offset : pointerStack
-                          target = B.drop offset inp
-                      case runSGet (getDomain' sep1 newStack) target of
+                      case runSGet (getDomain' sep1 lim newStack) target of
                           Left (DecodeError err) -> fail err
                           Left err               -> fail $ show err
                           Right o  -> push pos (fst o) >> return (fst o)
@@ -255,7 +259,7 @@ getDomain' sep1 pointerStack = do
         _ | isExtLabel c -> return ""
         _ -> do
             hs <- getNByteString n
-            ds <- getDomain' '.' pointerStack
+            ds <- getDomain' '.' lim pointerStack
             let dom = case ds of -- avoid trailing ".."
                     "." -> hs `BS.append` "."
                     _   -> hs `BS.append` BS.singleton sep1 `BS.append` ds
--- a/Network/DNS/Decode/Internal.hs
+++ b/Network/DNS/Decode/Internal.hs
@@ -220,18 +220,21 @@ getOData opc len = UnknownOData opc <$> getNByteString len
 getDomain :: SGet Domain
 getDomain = do
     lim <- B.length <$> getInput
-    getDomain' '.' lim []
+    getDomain' '.' lim 0
 
 getMailbox :: SGet Mailbox
 getMailbox = do
     lim <- B.length <$> getInput
-    getDomain' '@' lim []
+    getDomain' '@' lim 0
 
 -- | Get a domain name, using sep1 as the separate between the 1st and 2nd
 -- label.  Subsequent labels (and always the trailing label) are terminated
 -- with a ".".
-getDomain' :: Char -> Int -> [Int] -> SGet ByteString
-getDomain' sep1 lim pointerStack = do
+getDomain' :: Char -> Int -> Int -> SGet ByteString
+getDomain' sep1 lim loopcnt
+  -- 127 is the logical limitation of pointers.
+  | loopcnt >= 127 = fail "too deep pointers"
+  | otherwise      = do
     pos <- getPosition
     c <- getInt8
     let n = getValue c
@@ -245,11 +248,9 @@ getDomain' sep1 lim pointerStack = do
             mo <- pop offset
             case mo of
                 Nothing
-                  | offset `elem` pointerStack -> return "" -- pointer loop
                   | otherwise                  -> do
                       target <- B.drop offset <$> getInput
-                      let newStack = offset : pointerStack
-                      case runSGet (getDomain' sep1 lim newStack) target of
+                      case runSGet (getDomain' sep1 lim (loopcnt + 1)) target of
                           Left (DecodeError err) -> fail err
                           Left err               -> fail $ show err
                           Right o  -> push pos (fst o) >> return (fst o)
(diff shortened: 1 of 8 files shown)


kazu-yamamoto:
Ping @benclifford and @vdukhovni

Network/DNS/Decode/Internal.hs
```diff
@@ -240,9 +239,12 @@ getDomain' sep1 = do
             let offset = n * 256 + d
             mo <- pop offset
             case mo of
-                Nothing -> fail $ "getDomain: " ++ show offset
-                -- A pointer may refer to another pointer.
-                -- So, register this position for the domain.
+                Nothing -> do
+                    inp <- getInput
+                    case runSGet getDomain (B.drop offset inp) of
```
vdukhovni: In the very unlikely case that an SOA `mrname` is a pointer to another domain, should the `getDomain` on this line be `(getDomain' sep1)`?

vdukhovni: Also what prevents infinite pointer chasing recursion in maliciously constructed reply packets?

Network/DNS/IO.hs
```diff
@@ -26,51 +27,78 @@ module Network.DNS.IO (
...
-sink :: ConduitM ByteString o IO (DNSMessage, PState)
-sink = sinkParser $ ST.runStateT getResponse initialState
-
 -- | Receiving DNS data from 'Socket' and parse it.
 
 receive :: Socket -> IO DNSMessage
-receive sock = fst <$> runConduit (sourceSocket sock .| sink)
+receive sock = do
+    bs <- recv sock 4096 `E.catch` \e -> E.throwIO $ NetworkFailure e
+    case decode bs of
```
vdukhovni: The hard-coded `4096` here at least deserves a comment.  We're expecting DNS replies to not exceed this length presumably because that's our advertised EDNS UDP buffer size?  It would probably do no harm to use a larger packet size here, the most we get is a single UDP datagram of 65535 bytes, perhaps we should attempt to read the whole thing whatever the size?

vdukhovni:
A couple of initial comments, I'll look at the rest soon...

Network/DNS/IO.hs
benclifford: @vdukhovni in principle, ipv6 jumbograms can be bigger than 65535 bytes, but I don't know the technicalities of whether this would happen with DNS.

vdukhovni: The ironic thing is that it is precisely with IPv6 that DNS is observed to have the most difficulties with dropped IP fragments. There are reports of lost data with sizes as small as >1280 bytes.  I don't think we need to worry about larger responses and jumbo IPv6, just 65535 (largest possible IPv4 UDP size) is enough.  Of course if can be sure that the EDNS(0) advertisement matches the limit here, we might also be OK.

kazu-yamamoto: I would like to set:


```
maxUdpSize = 16384
minUdpSize = 512
```

then change `fromEDNS0` ensuring UDP size is one between them and use `maxUdpSize` in `receive`.

Network/DNS/Decode/Internal.hs
kazu-yamamoto: > In the very unlikely case that an SOA `mrname` is a pointer to another domain, should the `getDomain` on this line be `(getDomain' sep1)`?

Right. Fixed.

kazu-yamamoto: > Also what prevents infinite pointer chasing recursion in maliciously constructed reply packets?

c27ef782013e53cdfda055b3af29aa4f7fc40190 should fix this.

Network/DNS/Decode/Internal.hs
```diff
@@ -239,19 +239,23 @@ getDomain' sep1 = do
...
-                    case runSGet (getDomain' sep1) (B.drop offset inp) of
-                      Left (DecodeError err) -> fail err
-                      Left err               -> fail $ show err
-                      Right o  -> push pos (fst o) >> return (fst o)
+                Nothing
+                  | offset `elem` pointerStack -> return "" -- pointer loop
+                  | otherwise                  -> do
+                      inp <- getInput
+                      let newStack = offset : pointerStack
+                          target = B.drop offset inp
```
vdukhovni: Should we check that `offset` is less than length of `inp`?  Or just rely on `Data.ByteString.drop` returning an empty string, from which we'll fail to parse a domain (no length/ptr field for the first label)?

As for loop detection we could instead limit the pointer recursion depth. A sensibly compressed name should never have a pointer directly to another pointer, it should then just point at that pointer's target. So the target of any pointer should be a `real` label. Since it takes at least two bytes to encode a label, and wire-form domain names are at most 255 bytes, with thus, at most every other label a non-pointer we can stop at 127 pointers:
```
->(l1.(->a2.(->a3.(...(->a127.)....))))
```
So it is probably cheaper to limit the recursion depth to 127 than to look for cycles i the visited offsets...
Of course cycle detection works too, but it feels needlessly costly (am I doing premature optimization?)

vdukhovni:
Overall, this seems to do what's expected: I took out the parser error work-around and nothing bad happened:
```
aghering.tplinkdns.com. IN A 83.86.241.197 ; NoError AD=0
mail.aghering.com. IN AAAA ? ; DecodeError "Failed reading: IPv6 addresses must be 16 bytes long" AD=0
_25._tcp.mail.aghering.com. IN TLSA ? ; NODATA AD=0
```
Previously, I had an unhandled exception in the `AAAA` lookup and never got to reporting the TLSA status (inferred not looked up, but that's not important).

Network/DNS/Decode/Internal.hs
kazu-yamamoto: Both are done!

Network/DNS/Decode/Internal.hs
```diff
@@ -220,18 +220,21 @@ getOData opc len = UnknownOData opc <$> getNByteString len
...
 
 -- | Get a domain name, using sep1 as the separate between the 1st and 2nd
 -- label.  Subsequent labels (and always the trailing label) are terminated
 -- with a ".".
-getDomain' :: Char -> Int -> [Int] -> SGet ByteString
-getDomain' sep1 lim pointerStack = do
+getDomain' :: Char -> Int -> Int -> SGet ByteString
+getDomain' sep1 lim loopcnt
+  -- 127 is the logical limitation of pointers.
+  | loopcnt >= 127 = fail "too deep pointers"
```
vdukhovni: I would recommend s/too deep pointers/pointer recursion limit exceeded/

kazu-yamamoto: Done.

kazu-yamamoto:
I have merged this PR.
Thank you for your reviews.
I will release a new version soon.

PR #112, opened 2018-05-17
Drop unexpected UDP answers

vdukhovni:
Occassionally, when a client gives on a query and closes its UDP
socket, the nameserver nevertheless eventually replies.  When that
happens, the UDP socket in question may now be in use for a different
query.  When the unexpected stale answer arrives, we should just
drop it, and continue waiting for the right answer.  Otherwise, we
may end up with spurious sequence number mismatch errors (observed
in practice under heavy load, with thousands of DNS queries per
second).

We check both the sequence number and the question.  For this the
question domain needs to be in standard form, with a trailing '.'.
We might have checked the question class, but questions are at
present implicitly in class "IN".  Sprinkled in some comments in
case that ever changes.

Turned the question into a singleton list early, simplifying
downsream code.

### DIFF
--- a/Network/DNS/Decode/Internal.hs
+++ b/Network/DNS/Decode/Internal.hs
@@ -80,6 +80,9 @@ getTYPE = toTYPE <$> get16
 getOptCode :: SGet OptCode
 getOptCode = toOptCode <$> get16
 
+-- XXX: Include the class when implemented, or otherwise perhaps check the
+-- implicit assumption that the class is classIN.
+--
 getQuery :: SGet Question
 getQuery = Question <$> getDomain
                     <*> getTYPE
--- a/Network/DNS/Encode.hs
+++ b/Network/DNS/Encode.hs
@@ -98,10 +98,12 @@ putDNSFlags DNSFlags{..} = put16 word
 
     word = execState st 0
 
+-- XXX: Use question class when implemented
+--
 putQuestion :: Question -> SPut
 putQuestion Question{..} = putDomain qname
                            <> put16 (fromTYPE qtype)
-                           <> put16 1
+                           <> put16 classIN
 
 putResourceRecord :: ResourceRecord -> SPut
 putResourceRecord ResourceRecord{..} = mconcat [
--- a/Network/DNS/Transport.hs
+++ b/Network/DNS/Transport.hs
@@ -19,11 +19,12 @@ import Network.DNS.Imports
 import Network.DNS.Types
 import Network.DNS.Types.Internal
 
--- | Check response for a matching identifier.  If we ever do pipelined TCP,
--- we'll also need to match the QNAME, CLASS and QTYPE.  See:
+-- | Check response for a matching identifier and question.  If we ever do
+-- pipelined TCP, we'll need to handle out of order responses.  See:
 -- https://tools.ietf.org/html/rfc7766#section-7
-checkResp :: Question -> Identifier -> DNSMessage -> Bool
-checkResp _ seqno resp = identifier (header resp) == seqno
+checkResp :: [Question] -> Identifier -> DNSMessage -> Bool
+checkResp q seqno resp =
+   (identifier (header resp) == seqno) && (q == (question resp))
 
 ----------------------------------------------------------------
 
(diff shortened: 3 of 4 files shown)


dudebout:
Checking the question section is actually problematic in the TCP case. If there is a mismatch in the question, you get the SequenceNumberMismatch error.

1. As it stands, i.e. without pipelined TCP, does checking the question add something?
2. Does checking the question in the UDP case actually add something? Is it possible to have the same transaction ID but a different question on the same socket?

If the answer to these questions is no, I think we should remove the check for the question section.

vdukhovni:
> @dudebout Does checking the question in the UDP case actually add something? Is it possible to have the same transaction ID but a different question on the same socket?

Yes, under high UDP query load stale answers to queries timed out by the client were delivered to the client after the client closed the original socket and opened many new ones, including finally one with the same port number as the delayed answer from the server.  This happens when the resolver keeps requests in the queue longer than the client is willing to wait for an answer.

As for TCP, please explain the circumstances under which you're observing a problem? `Network.DNS` isn't presently doing pipelining, so the question should always match, and checking is harmless.

vdukhovni:
Note also that when the DNS server is remote, not checking the question increases opportunities for cache poisoning by an off-path attacker. Crypto-random UDP ports, plus crypto-random query ids plus checks that the question matches help to harden the query against forged answers. We might some day even support "Use of Bit 0x20 in DNS Labels to Improve Transaction Identity" (https://tools.ietf.org/html/draft-vixie-dnsext-dns0x20-00).  My resolver is "0x20 compatible":
```
$ dig +qu WwW.iETf.Org
; <<>> DiG 9.12.3-P1 <<>> +qu WwW.iETf.Org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 57920
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 8192
;; QUESTION SECTION:
;WwW.iETf.Org.                  IN      A

;; ANSWER SECTION:
WwW.iETf.Org.           67      IN      CNAME   www.ietf.org.cdn.cloudflare.net.
www.ietf.org.cdn.cloudflare.net. 127 IN A       104.20.0.85
www.ietf.org.cdn.cloudflare.net. 127 IN A       104.20.1.85

;; Query time: 39 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Sun Feb 24 15:17:24 EST 2019
;; MSG SIZE  rcvd: 118
```

dudebout:
> Yes, under high UDP query load stale answers to queries timed out by the client were delivered to the client after the client closed the original socket and opened many new ones, including finally one with the same port number as the delayed answer from the server. This happens when the resolver keeps requests in the queue longer than the client is willing to wait for an answer.

In this case would the transaction ID, not be sufficient to determine that the same socket has been reused?

On the TCP front, I encountered the issue while working on adding AXFR support: https://github.com/kazu-yamamoto/dns/pull/134. AXFR uses a TCP connection but may or may not have the question in the reply: https://cr.yp.to/djbdns/axfr-notes.html gives some examples where bind and the rest of the world differ.

For the sake of discussion, I have created the PR with what I had, where I have removed checking the question. I'd be happy to improve the PR to reintroduce that check if requested.

vdukhovni:
> @dudebout In this case would the transaction ID, not be sufficient to determine that the same socket has been reused?

No, the idea is to distinguish between `sequence number mismatch` and `stale answer to a different question`.

As for TCP and AXFR, AXFR is a very different beast. I'd prefer if the check could be made optional just in that case. Indeed AXFR returns a stream of answers enclosed in a pair of matching SOA RRs, so the handling of AXFR is already rather special. It might make sense to `fork-lift` rather than re-use the existing TCP query code to create specialized AXFR code.  When code duplication reduces complexity by not overloading too many use-cases into a single interface, code duplication is a feature not a bug.

dudebout:
> No, the idea is to distinguish between `sequence number mismatch` and `stale answer to a different question`.

In my first attempt, I had added a new error data constructor `QuestionMismatch` and split the two cases. It made sense in the TCP case, where I could return the user two different error messages. However, for the UDP case, both cases were ignored. I started the discussion here to better understand the rationale.

> It might make sense to `fork-lift` rather than re-use the existing TCP query code to create specialized AXFR code.

I also had a version where I introduced `axfrLookup` as a modified copy of `tcpLookup`. It was so similar to `tcpLookup` that I decided against it for the time being, but once again could rework the PR if requested. Let me know if you think #134 increases complexity in that respect.

PR #113, opened 2018-06-21
Add NOTIFY and UPDATE opcodes

themattchan:
No further changes are required to support these extensions since they reuse the DNS message format.


kazu-yamamoto:
If you would like to have a new version, please let me know.

PR #116, opened 2018-10-04
Make use of the RD and CD bits configurable

vdukhovni:
This can be done at the resolver level by setting default values for
new ResolvConf fields.  This makes it possible to request unvalidated
data from validating DNSSEC resolvers (by setting the CD bit) or
to query authoritative servers (by clearing the RD bit, since many
authoritative servers will refuse queries with the RD bit set, even
when authoritative for the query domain).

Alternatively, the new "lookupRaw'" function takes a "QueryFlags"
parameter which is a "newtype" around DNFlags with a hidden
constructor.  The "mkQueryFlags" function returns a "QueryFlags"
record in which only query-related flags can be modified.  The
"fromQueryFlags" accessor gives us back the contructed "DNSFlags".

With these changes one can configure per-resolver defaults for the
RD/AD/CD bits, or else choose settings on the fly for each lookup
with "lookupRaw'".  (This makes "lookupRawAD" somewhat obsolete,
we might deprecate it at some point in the future, but probably
best to leave it be, it is simpler if one just wants the AD bit).

### DIFF
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -52,8 +52,19 @@ module Network.DNS.Types (
   -- ** DNS Header
   , DNSHeader (..)
   , Identifier
+  -- *** DNS flags
+  , DNSFlags (..)
   , QorR (..)
+  , defaultDNSFlags
+  , QueryFlags
+  , queryDNSFlags
+  , rdBit
+  , adBit
+  , cdBit
+  -- **** OPCODE and RCODE
   , OPCODE (..)
+  , fromOPCODE
+  , toOPCODE
   , RCODE (
     NoErr
   , FormatErr
@@ -68,17 +79,6 @@ module Network.DNS.Types (
   , NotZone
   , BadOpt
   )
-  -- *** DNS flags
-  , DNSFlags (..)
-  , defaultDNSFlags
-  , QueryFlags
-  , queryDNSFlags
-  , rdBit
-  , adBit
-  , cdBit
-  -- **** OPCODE and RCODE
-  , fromOPCODE
-  , toOPCODE
   , fromRCODE
   , toRCODE
   , fromRCODEforHeader
@@ -443,6 +443,8 @@ defaultDNSFlags = DNSFlags
 -- yield all possible combinations of "set", "clear" and "reset" (to default)
 -- for each of the bits.
 --
+-- ==== __Example__
+--
 -- >>> :{
 -- let setrd = rdBit (Just True)
 --     setad = adBit (Just True)
@@ -542,6 +544,8 @@ data OPCODE
   | OP_UPDATE -- ^ An update request (RFC2136)
   deriving (Eq, Show, Enum, Bounded)
 
+-- | Convert a 16-bit DNS OPCODE number to its internal representation
+--
 toOPCODE :: Word16 -> Maybe OPCODE
 toOPCODE i = case i of
   0 -> Just OP_STD
@@ -551,6 +555,9 @@ toOPCODE i = case i of
   5 -> Just OP_UPDATE
   _ -> Nothing
 
+-- | Convert the internal representation of a DNS OPCODE to its 16-bit numeric
+-- value.
+--
 fromOPCODE :: OPCODE -> Word16
 fromOPCODE OP_STD    = 0
 fromOPCODE OP_INV    = 1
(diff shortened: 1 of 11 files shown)


vdukhovni:
The new commit is the same code, but corrects some typos in the commit message.

Network/DNS/Types.hs
```diff
@@ -405,9 +411,66 @@ data DNSFlags = DNSFlags {
...
+-- | Return 'DNSFlags' appropriate for a query with the specified values of
+-- the RD, AD and CD bits.  This can be used to update the return value of
+-- 'defaultQuery' with the desired settings of those bits:
+--
+-- >>> :{
+-- let hdr = header defaultQuery
+--     qry = defaultQuery {
+--     header = hdr {
+--              flags = fromQueryFlags $
+--                      mkQueryFlags (RDbit True) (ADbit True) (CDbit False)
```
kazu-yamamoto: I think that `defaultQueryFlags` should be defined somewhere and used here: `fromQueryFlags defaultQueryFlags`.

Network/DNS/Types.hs
```diff
@@ -405,9 +411,66 @@ data DNSFlags = DNSFlags {
...
+--              }
+--          }
+-- :}
+--
+mkQueryFlags :: RDbit -> ADbit -> CDbit -> QueryFlags
```
kazu-yamamoto: `mkQueryFlags` should use `defaultQueryFlags`, too.

vdukhovni:
Thanks for the feedback, I'm reworking the query flags interface, updated patch soon...

vdukhovni:
New version force-pushed. This is not an incremental change, but rather a replacement. Sorry about any duplication of work in the review, but it should all be fairly simple.

There are now two commits, the second of which is just some documentation tweaks I thought might be useful, since I happened to be looking at those docs when making my changes.

vdukhovni:
One more force-push to deal with the (Semigroup => Monoid) constraint in GHC 8.4+

vdukhovni:
Finally all the Semigroup/Monoid issues sorted out portably across the CI builds. Ready for review.

vdukhovni:
@kazu-yamamoto Any feedback on the updated implementation?

vdukhovni:
The `FlagOp` change is good. No issues. The `Maybe (Maybe Bool)` variant was a "cute" way to exercise the `Alternative` instance of `Maybe`, but though logically equivalent, was a less clear interface.

As for dropping `lookupRawAD`, `composeQuery` and `composeQueryAD`, I am skeptical that anyone other than you and I used these.  So no objections from me, and perhaps nobody else will be affected, but I really don't know what the impact on others might be. Google does not seem to find any other users. Your call.

Changing the interface of `encodeQuestions` also looks pretty safe to me, that's a fairly low-level function, that most users are unlikely to use.

The change that looks plausibly risky to downstream consumers is in `lookupRaw`.  While I've not managed to Google real code that uses it outside Network.DNS itself, there is at least this article that has a `lookupRaw`

example: http://www.shakthimaan.com/posts/2015/12/30/haskell-network-programming/news.html

So I'm not sure that change is prudent, but again you may have a better sense how much downstream breakage if any to expect.

vdukhovni:
Specifically, re `lookupRaw` vs `lookupRaw'`, the first always uses the query flags from the resolver, while the latter takes an additional argument with per-lookup flag overrides.  While one can of course use `lookupRaw rlv dom typ mempty` rather than `lookupRaw rlv dom typ` that does seem a bit tedious, and exposes a design feature (flag overrides as a monoid) that users who don't want per-query overrides might avoid depending on in their code.

If we provide `lookupRaw` as the function that uses just the resolver settings and `lookupRawWithFlags` (or something like that) that also takes a "QueryFlags" parameter, then most users can stick with the simpler interface, and avoid the `mempty` which does more than save keystrokes, it avoids a dependency on this new feature in their code.  So I guess I am leaning towards replacing `lookupRawAD` with the more general version, but still having two separate functions `lookupRaw` and `lookupRawFlags` or `lookupRawEx`,  or `lookupRaw'`, ... whatever name is most appropriate.

vdukhovni:
Bottom line I think we should have two functions for "lookupRaw", one that takes the extra customization argument, and one that offers the default behaviour.

kazu-yamamoto:
`lookupRawWithFlags` is defined.
Are we ready for merging?

vdukhovni:
Yes, basically ready.  I suggested what I think is somewhat more idiomatic English in the haddock text for `{rd,ac,cd}Flag'. I assume tests are passing?  I've not yet run this code, though I don't expect any problems.

kazu-yamamoto:
> I assume tests are passing?

Yes.

kazu-yamamoto:
Merged.

PR #118, opened 2018-10-11
adding resolvNumericHostOnly (#115)

### LINKED ISSUE #115: makeResolvSeed throws an error when /etc/resolv.conf contains a non-numeric nameserver
Today I encountered this error message in one of my applications that has `dns` as a transitive dependency:

```
werewolf-xmpp: Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG,AI_NUMERICHOST,AI_PASSIVE], addrFamily = AF_UNSPEC, addrSocketType = Datagram, addrProtocol = 0, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "localhost", service name: Just "domain"): does not exist (Name or service not known)
```

As it turns out, this seems to happen, because my `/etc/resolv.conf` contains an entry like this:

```
nameserver localhost
```

As it turns out calls to `makeResolvSeed defaultResolvConf` throw the exception mentioned above when `/etc/resolv.conf` contains a nameserver with a non-numeric name (such as the `localhost` above). This is because `getAddrInfo` is given the `AI_NUMERICHOST` option, when `localhost` is clearly not a numeric host.

Now, I know it might seem reasonable to assume that `/etc/resolv.conf` may only contain numeric addresses, because you can't do dns lookups before having read that file, but the assumption doesn't actually hold up, because hostnames that can be resolved via `/etc/hosts` (or some other strange way people might have configured in their libnss) work just fine in `/etc/resolv.conf`. This `dns` library is really the only piece of software that seems to have a problem with this setup.

Of course I can just work around this by replacing the `localhost` with `127.0.0.1` or `::1`, but IMO this should really be something that should just work out of the box.

kazu-yamamoto:
@Kritzefitz This should fix #115. Would you review this?

### DIFF
--- a/Network/DNS/Resolver.hs
+++ b/Network/DNS/Resolver.hs
@@ -15,6 +15,7 @@ module Network.DNS.Resolver (
   , resolvConcurrent
   , resolvCache
   , resolvQueryFlags
+  , resolvNumericHostOnly
   -- ** Specifying DNS servers
   , FileOrNumericHost(..)
   -- ** Configuring cache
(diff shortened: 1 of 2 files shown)


kazu-yamamoto:
No merging.

PR #123, opened 2018-10-15
For FlagOp and QueryFlags, the right value should win.

kazu-yamamoto:
@vdukhovni It seems to me that it is more natural that right values win. What do you think?

### DIFF
--- a/Network/DNS/Transport.hs
+++ b/Network/DNS/Transport.hs
@@ -80,7 +80,7 @@ resolve dom typ rlv qfl rcv
     seed    = resolvseed rlv
     nss     = NE.toList $ nameservers seed
     onlyOne = length nss == 1
-    fl      = qfl <> resolvQueryFlags (resolvconf $ resolvseed rlv)
+    fl      = resolvQueryFlags (resolvconf $ resolvseed rlv) <> qfl
 
     conf       = resolvconf seed
     concurrent = resolvConcurrent conf
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -439,15 +439,15 @@ defaultDNSFlags = DNSFlags
          }
 
 -- | Flag operations. This is an instance of 'Monoid'.
--- If they are used with '(<>)', the left value wins.
+-- If they are used with '(<>)', the right value wins.
 --
 -- >>> mempty :: FlagOp
 -- FlagKeep
--- >>> FlagSet <> mempty
+-- >>> mempty <> FlagSet
 -- FlagSet
--- >>> FlagClear <> FlagSet <> mempty
+-- >>> mempty <> FlagSet <> FlagClear
 -- FlagClear
--- >>> FlagReset <> FlagClear <> FlagSet <> mempty
+-- >>> mempty <> FlagSet <> FlagClear <> FlagReset
 -- FlagReset
 data FlagOp = FlagSet   -- ^ Flag is set
             | FlagClear -- ^ Flag is unset
@@ -463,8 +463,8 @@ data FlagOp = FlagSet   -- ^ Flag is set
 -- True
 --
 instance Sem.Semigroup FlagOp where
-    FlagKeep <> op = op
-    op       <> _  = op
+    op <> FlagKeep = op
+    _  <> op       = op
 
 instance Monoid FlagOp where
     mempty = FlagKeep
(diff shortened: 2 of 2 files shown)


vdukhovni:
Another possibly relevant observation is that `mconcat` uses `foldr` and so we get:
```
import Data.Semigroup
import Data.Monoid

data Foo = Foo String deriving (Show)

-- Deliberately non-associative
instance Semigroup Foo where
    (Foo a) <> (Foo b) = Foo $ "(" ++ a ++ ", " ++ b ++ ")"

instance Monoid Foo where
    mempty = Foo $ ""

main :: IO ()
main = print $
    mconcat [Foo "a", Foo "b", Foo "c"]
```
printing `Foo "(a, (b, (c, )))"`. Again not sure that changes anything...

kazu-yamamoto:
Right associative operators are friendly *lists* since sometime we can skip the evaluation of *tail* parts. Also, (++) is right associative since it can avoid creating unnecessary itermediate lists:

```haskell
"foo" ++ "bar" ++ "baz" = "foo" ++ ("bar" ++ "baz") -> "foo" ++ "barbaz" -> "foobarbaz"
```
In this case, "baz" and "barbaz" are shared.

`(<>)` is defined as right associative because it is generization of (++).

But `FlagOp` is not a list. We can freely choose left-win or right-win. Also, you may be surprized that Applicative.(<|>) is defined as left associative.

We uses `FlagOp` as *sequence*, so I prefer right-win.

kazu-yamamoto:
@snoyberg We are disussing a semigroup (or monoid) instance of *configuration* data, either left-win or right-win.

Left-win:

```Haskell
>>> FlagReset <> FlagClear <> FlagSet <> mempty
FlagReset
```

Right-win:

```Haskell
>> mempty <> FlagSet <> FlagClear <> FlagReset
FlagReset
```

In this specific case, I feel that right-win is more natural. Any suggestions?

vdukhovni:
Does https://en.wikibooks.org/wiki/Haskell/Monoids#Uses help?

snoyberg:
I've personally gone in the direction of setter functions with opaque types, since this makes it easier to extend types in the future. This naturally leads to left-bias. Instead of a `Semigroup` or `Monoid` instance, you end up with function composition, and instead of `mempty` you have a monomorphic default settings type. For example:

```haskell
setFlagSet $ setFlagClear $ setFlagReset defaultDnsSettings
```

would be the same as

```haskell
setFlagSet defaultDnsSettings
```

If you want to go for the `Monoid` instance, I'd recommend left biased instead, since it matches both functions and common data structures like `Map`.

vdukhovni:
So while I still don't know whether left or right is inherently more natural, with "left" as my initial guess also by analogy with function composition (endomorphisms), this discussion does give me time to consider another aspect of the API.  Right now all the flags of interest (RD, AD, CD) are in the `DNSFlags` part of the `DNSMessage` header, and so our monoid ultimately leads to a map producing the relevant query header field:
```
queryDNSFlags :: QueryFlags -> DNSFlags
```
It occurs to me that this may prove too restrictive, because, for example, the "DO" bit is not in the message header, but is carried in the OPT RR pseudo-header.  So if some day we decide to also support setting the "DO" bit in queries, the operation in question should perhaps be:
```
queryDNSFlags :: QueryFlags -> DNSMessage
```
that is, the monoid should perhaps induce a transformation of `defaultQuery` (a skeletal query `DNSMessage`) rather than of `defaultDNSFlags` (the `DNSFlags` component of just the message header).

That is, we're allowing users to configure composable transformations of the query message skeleton, which we then complete by adding the question and the EDNS0 size, ...  But it then means that we'd need to be able to set the EDNS0 size without clobbering the "DO" bit, ... that is, again in a way that composes with whatever we allow to be configurable at the resolver and lookup layers.

This does not mean we have to rush to implement DO bit support now.  Nor does it even mean that the monoid could not be extended with more bits later, but given that the transformation function is exposed to users who want to create their own messages, we should probably decide whether the it applies to just message header flags, or to complete messages.

kazu-yamamoto:
@snoyberg I'm a fan of `Monoid` but not a fan of `Monoid` in the case of configurations. Since I'm a WAI guy, I like the `defaultDnsSettings` approach. :-) Actually, I proposed this: https://github.com/kazu-yamamoto/dns/pull/120

But Viktor likes the monoid approach.

Anyway, let's take the left-biased.

@snoyberg Thank you for your suggestion.

vdukhovni:
The reason I like the monoid approach is that the user specifies just the settings which are generated as combinations of three sets distinct per-flag operators, and does not immediately apply them to a default, that happens later. The reason that's important is that transformations compose, but results do not.  And with transformations, we can specify some at the resolver, and others per query.  These are naturally a monoid, though we arguably don't have to expose that fact to the user.  Still:
```
resolvQueryFlags = rdFlag FlagSet <> adFlag <> FlagClear <> cdFlag FlagSet
```
looks like a sensible UI to me.  Since these are not actually functions, but just represent intended side-effects, we don't expose arbitrary mappings of the header flag field to itself, but expose just a controlled subset.  This too is IMHO a feature.  But as for left bs. right, I really have no idea which is better, I just followed a few precedents that I thought were sort of close, and perhaps was not too far off the mark.

The remaining issue is now not the monoid itself, but its "interpretation" via `queryDNSFlags`. What do you think of my suggestion of making it return a `DNSMessage` rather than just the headers, allowing future expansion to include the "DO" bit or other things that might later be added to EDNS. Perhaps support for DNS cookies?  Etc.  While at the resolver layer those other things could just use separate controls, on a per query basis the API is more constrained, and extending the monoid leaves room for further extensions, but the side effects would now be on the rest of the message.

kazu-yamamoto:
@vdukhovni I agree with you and am now thinking the last issue which you pointed out.

kazu-yamamoto:
Close this in favor of #124.

PR #124, opened 2018-10-16
exporting makeEmptyQuery instead of queryDNSFlags.

kazu-yamamoto:
@vdukhovni Please review this to resolve the issue in #123. I hope you like this.

### DIFF
--- a/Network/DNS/IO.hs
+++ b/Network/DNS/IO.hs
@@ -143,14 +143,14 @@ encodeQuestions idt qs adds fs = encode $ makeQuery idt qs adds fs
 
 -- | Composing a response from IPv4 addresses
 responseA :: Identifier -> Question -> [IPv4] -> DNSMessage
-responseA idt q ips = makeResponse idt [q] as
+responseA idt q ips = makeResponse idt q as
   where
     dom = qname q
     as  = ResourceRecord dom A classIN 300 . RD_A <$> ips
 
 -- | Composing a response from IPv6 addresses
 responseAAAA :: Identifier -> Question -> [IPv6] -> DNSMessage
-responseAAAA idt q ips = makeResponse idt [q] as
+responseAAAA idt q ips = makeResponse idt q as
   where
     dom = qname q
     as  = ResourceRecord dom AAAA classIN 300 . RD_AAAA <$> ips
--- a/Network/DNS/IO.hs
+++ b/Network/DNS/IO.hs
@@ -9,6 +9,7 @@ module Network.DNS.IO (
   , send
   , sendVC
     -- ** Encoding queries for transmission
+  , encodeQuestion
   , encodeQuestions
     -- ** Creating query response messages
   , responseA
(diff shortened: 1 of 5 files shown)


kazu-yamamoto:
@vdukhovni I added two more commits. Would you review this PR again?

I don't know which is better for the signature of `makeResponse`:

```
makeResponse :: Identifier -> [Question] -> [ResourceRecord] -> DNSMessage
```

or

```
makeResponse :: Identifier -> Question -> [ResourceRecord] -> DNSMessage
```

vdukhovni:
I'm inclined to say the second:
```
makeResponse :: Identifier -> Question -> [ResourceRecord] -> DNSMessage
```
and perhaps then go further and implement:
```
encodeQuestion :: Identifier -> Question -> [ResourceRecord] -> QueryFlags -> ByteString
```
deprecating `encodeQuestions`

kazu-yamamoto:
OK. I have implemented them.

Would you check if the last commit is necessary.

vdukhovni:
Thanks!  Yes, while I am not opposed to the last commit, I think it perhaps adds more complexity than benefit. So if you're not entirely fond of it either, feel free to leave it out.

Looking at all the good work done up to this point, I am somewhat inclined to also reconsider `resolvEDNS`, this is a crude explicit RRset, which lacks fine-grained composability.  Now that the QueryFlags transform DNS messages, which could introduce some combinators that construct the EDNS OPT record from smaller building blocks, and fold these into the QueryFlags, whose settings in the resolver can be augmented with additional properties via lookupRaw.

But perhaps you're tired of continuing to explore the design space, and want to be done and ship?  We could stop here for now, but I thought I'd ask the question, in case there's interest to push the ideas through to their conclusion...

vdukhovni:
I can do these as a separate follow-up PR if you like? Either way, we could merge and close this one, and do any followups separately.

kazu-yamamoto:
> Thanks!  Yes, while I am not opposed to the last commit, I think it perhaps adds more complexity than benefit. So if you're not entirely fond of it either, feel free to leave it out.

I like this commit because `makeQuery` gets similar to `makeResponse`. So, I keep it.

kazu-yamamoto:
> Looking at all the good work done up to this point, I am somewhat inclined to also reconsider `resolvEDNS`, this is a crude explicit RRset, which lacks fine-grained composability.  Now that the QueryFlags transform DNS messages, we could introduce some combinators that construct the EDNS OPT record from smaller building blocks, and fold these into the QueryFlags, whose settings in the resolver can be augmented with additional properties via lookupRaw.

Yes, it's high time to do this.

> But perhaps you're tired of continuing to explore the design space, and want to be done and ship?  We could stop here for now, but I thought I'd ask the question, in case there's interest to push the ideas through to their conclusion...

I'm not tired. Rather, I'm interested in more robust APIs. Let's do it.

kazu-yamamoto:
> For example, we could introduce an `ednsUdpSize :: Word16 -> QueryFlags` combinator (perhaps renaming `QueryFlags` to `QueryOptions` in the process). 

I prefer to `QueryOptions`.

kazu-yamamoto:
> I can do these as a separate follow-up PR if you like? Either way, we could merge and close this one, and do any followups separately.

Yes. I will merge this PR as is and pass the batton to you. Please send a PR and let's discuss better query options based on the PR.

kazu-yamamoto:
Merged.

vdukhovni:
This PR is closed, but just FYI, I now have working code for the ideas discussed here, but one thing led to another, and I ended making various additional breaking changes, that won't affect users of the high-level APIs, but will affect users of the EDNS and low-level message processing API.

Before I re-organize this into a series of commits, I wanted to gauge your "enthusiasm" for further API cleanup. How much of the low-level interface that few users are likely to touch am I at liberty to change at this time?

* I am changing EDNS0 -> EDNS and adding a version field.  There could be an EDNS1 ... some day.
* I am adding more "known" EDNS options
* I am adding some more known RCODEs
* I am removing the overlap between the high 8 bits of the RCODE in the EDNS OPT record, and the low 4 bits in the message header.  They're now completely separate, and the high 8 bits are now a `Word8`.  A new function: `messageRCODE :: DNSMesage -> RCODE` computes the RCODE from the message.
* I am fixing the encoding of the ClientSubnet option which incorrectly generated the address prefix
* I am adding an `ECSGeneric` `OData` constructor, so that parsing of the ECS option can be a total function, this handles inputs where the address family is not IPv4(1) or IPv6(2).
* I am removing the `resolvEDNS` field of `resolvConf`, that's now handled via `resolvQueryOptions`
* I am removing `encodeQuestions`, it no longer easily fits the model, since EDNS via explicit additional RRs is gone, instead the QueryOptions generate the outgoing EDNS OPT RR.
* I am adding query option combinators to generate the various EDNS features.

Perhaps this will mean DNS 4.0 rather than DNS 3.1, though ultimately I expect most users to not notice...

kazu-yamamoto:
I'm very much enthusiastic and you can modify this library as you like! Your approach sounds great.

PR #125, opened 2018-10-21
Edns overhaul

vdukhovni:
The five commits in this pull request change a lot of code.  Each builds on the previous one, and the code should build and test after each step (1, 1 + 2, 1 + 2 + 3, ...).

It may be easier to review the commits one at a time, and see fewer changes at a given time, with better focus on a particular set of issues.

On the other hand the combined set of patches change some pieces of code multiple times.  So it is perhaps more "effiecient" to review everything at once, if it is not too much to take in.

If you'd prefer a series of 5 pull requests, we could start with commit 1, get that done, then review 2, ...

My advice is to clone the branch, and review on your own machine with "git diff -U20" (or however much context you prefer), and then go to github only to make comments on anything you find.

Once all these changes are done, we may be looking at DNS 4.0, though frankly I expect that most users stick to the high-level API, which has not changed.  Only folks playing directly with `DNSMessage`, `lookupRaw`, ... might notice some changes.

More remains to be done.  There should, for example, probably be a cache to remember which the nameserver (protocol + address) don't support EDNS, so that subsequent queries with the same resolver handle can avoid doing constant non-EDNS fallback.

The server side of the API (being a DNS server rather than client) is presently fairly anaemic. Not sure anyone really uses that, but more could perhaps be done to support replies in the context of a particular request, ensuring consistent handling of EDNS  between request and reply.  That said I have no present  need for DNS server support code in Haskell, and don't know whether anyone else does.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # 3.1.0
 
+* Breaking change: EDNS0 is renamed to EDNS.
 * Breaking change: lookupRawAD, composeQuery, composeQueryAD are removed.
 * A new API: lookupRawWithFlags, QueryFlags, FlagOp, rdFlag, adFlag, cdFlag,
   and resolvQueryFlags. [#116](https://github.com/kazu-yamamoto/dns/pull/116)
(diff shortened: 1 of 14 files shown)


kazu-yamamoto:
Merged.

PR #126, opened 2018-10-22
stop exporting HeaderControls and EdnsControls.

kazu-yamamoto:
I think that these exports are not necessary.

### DIFF
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -55,17 +55,13 @@ module Network.DNS.Types (
   , defaultQuery
   -- ** Query Controls
   , QueryControls
-  -- *** Header Controls
-  , HeaderControls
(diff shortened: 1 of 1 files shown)


vdukhovni:
Indeed the exports are not needed, but there may be a bit of the "haddock" commentary text for these that should be kept or moved? What do you think?

http://imrryr.org/~viktor/haskell/doc/dns-3.0.4/Network-DNS-Types.html#g:8
http://imrryr.org/~viktor/haskell/doc/dns-3.0.4/Network-DNS-Types.html#g:9

vdukhovni:
Do you want me to move the docs, or will you do it as part of this PR?  I agree consolidation is probably best.

vdukhovni:
Done.

PR #127, opened 2018-10-24
switching time to hourglass.

kazu-yamamoto:
Let's use `hourglass` instead of `time`.


kazu-yamamoto:
Merged. Thank you for your review!

PR #128, opened 2018-10-26
RRsig and safe decode

vdukhovni:
[ Since a file got renamed and replaced, individual commits will be much easier to review than the final result ]

- Safety: Both 'decode' and 'decodeAt' must now consume exactly
  the complete input buffer or a DecodeError is returned.

  * The same applies to each complete message with 'decodeMany'
    and 'decodeManyAt'.  Any final encoded message segment at the
    end of the input buffer is still returned as the second
    element of the result pair.

- Bugfix: fixed incorrect decoding of TXT records, and corrected
  the associated test.

- Cleanup: More precise control over decoder error messages via
  'failSGet', which avoids the unhelpful Attoparsec "Failure
  reading: " error prefix.

- Cleanup: Simplified loop detection in name decompression, making
  use of a monotone strictly decreasing limit on valid "pointer"
  targets.

- Breaking change: In the "Decode" module, expose only the
  decode{,Many}{,At} functions.  The rest of the "Decode" module's
  functions are now internal, exposed only for testing.  These
  include:

  * decodeDNSHeader
  * decodeDNSFlags
  * decodeResourceRecord
  * decodeDomain
  * decodeMailbox

- Cleanup: Reworked Decode module structure:

    * Moved Decode.Internal to Decode.Parsers

    * Created a new Decode.Internal which is now exposed, and
      moved some functions there from Decode which are only
      exposed for testing, since they could not reliably be used
      except as part of decoding a full message.

- Feature: RRSIG support, we can now encode, decode and show RRSIG
  records.  This uses the new 'decodeAt' and 'decodeManyAt' API.

- New API: 'decodeAt' and 'decodeManyAt' make it possible for
  the decoder to get the current time, in order to decode some
  RR types (like RRSIG) whose full meaning is time-dependent.


vdukhovni:
OK, this batch is ready for review.

The only other major deficit with 4.0.0 is that `(decode $ encodeDomain $ domain)` does not round-trip when domains have labels with special characters (e.g. "."). We kludge around this with Mailbox for SOA, but the issue is more general. For full compliance with the specs we need to implement support `\nnn` decimal escapes, which would encode spaces, dots, and various non-ASCII characters.

Finally, I can contribute IDNA code for the canonicalization module...

Network/DNS/IO.hs
```diff
@@ -45,7 +48,8 @@ receive :: Socket -> IO DNSMessage
 receive sock = do
     let bufsiz = fromIntegral maxUdpSize
     bs <- recv sock bufsiz `E.catch` \e -> E.throwIO $ NetworkFailure e
-    case decode bs of
+    Elapsed (Seconds now) <- timeCurrent
+    case decodeAt (fromIntegral now) bs of
```
kazu-yamamoto: `fromIntegral` is not necessary.

Network/DNS/IO.hs
```diff
@@ -57,7 +61,8 @@ receiveVC :: Socket -> IO DNSMessage
 receiveVC sock = do
     len <- toLen <$> recvDNS sock 2
     bs <- recvDNS sock len
-    case decode bs of
+    Elapsed (Seconds now) <- timeCurrent
+    case decodeAt (fromIntegral now) bs of
```
kazu-yamamoto: `fromIntegral` is not necessary.

Network/DNS/IO.hs
vdukhovni: Yes, thanks. that'll be part of the review `fixup` commit.

Network/DNS/IO.hs
vdukhovni: Yes, thanks.

kazu-yamamoto:
Should `encodeResourceRecord` and others move to `Encode.Internal`?

vdukhovni:
Separately, I am also going to replace the "Text.printf" line for formatting the RRsig date with:
```
        T.unpack $ T.left 4 '0' year <> T.left 2 '0' mon <> T.left 2 '0' day
                <> T.left 2 '0'   hh <> T.left 2 '0'  mm <> T.left 2 '0' ss
```
Unless we decide to delete that amazingly clever (and blazingly fast) date conversion code and use something equivalent from "Hourglass". If we needed accurate calendar dates from before Sep 14 1752 we'd need corrections, but, fortunately, that's not an issue for DNS timestamps.

kazu-yamamoto:
If you want to export `Interanl` but want to hide the documentation, you can use:

```
{-# OPTIONS_HADDOCK hide #-}
```

kazu-yamamoto:
> Separately, I am also going to replace the "Text.printf" line for formatting the RRsig date with:

OK.

vdukhovni:
> @kazu-yamamoto `{-# OPTIONS_HADDOCK hide #-}`

Thanks for the tip.  Will do.

vdukhovni:
Pushed one more time with review feedback, for Travis/Appveyor to catch any unexpected new issues.

vdukhovni:
CI passed.  Pushed the changes.  Thanks for the review comments, and moral support for these changes.

PR #130, opened 2018-12-06
Using Pattern synonyms in GHC 8.0.

kazu-yamamoto:
Pattern synonyms can be exported without the "pattern" keyword from GHC 8.0.1
https://downloads.haskell.org/~ghc/8.0.1/docs/html/users_guide/glasgow_exts.html#patsyn-impexp

### DIFF
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -168,7 +168,7 @@ type Mailbox = ByteString
 
 ----------------------------------------------------------------
 
-#if __GLASGOW_HASKELL__ >= 802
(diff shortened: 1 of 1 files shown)


vdukhovni:
Sure, if you say so. :-)

Any idea why some of the travis builds failed to find some dependencies?

kazu-yamamoto:
> Sure, if you say so. :-)

Merged. Thanks.

> Any idea why some of the travis builds failed to find some dependencies?

No. I confirmed that this happens even for `master`.

PR #134, opened 2019-02-24
Add AXFR support

dudebout:
Add support zone transfer (AXFR request type)


Network/DNS/Lookup.hs
```diff
@@ -429,3 +430,26 @@ lookupSRV rlv dom = do
...
...eRecord {rrname = "sip.zonetransfer.me.", rrtype = TYPE35, rrclass = 1, rrttl = 3333, rdata = \# 59 000200030150074532552b7369702b215e2e2a24217369703a637573746f6d65722d73657276696365407a6f6e657472616e736665722e6d652100},ResourceRecord {rrname = "sqli.zonetransfer.me.", rrtype = TXT, rrclass = 1, rrttl = 300, rdata = ' or 1=1 --},ResourceRecord {rrname = "sshock.zonetransfer.me.", rrtype = TXT, rrclass = 1, rrttl = 7200, rdata = () { :]}; echo ShellShocked},ResourceRecord {rrname = "staging.zonetransfer.me.", rrtype = CNAME, rrclass = 1, rrttl = 7200, rdata = www.sydneyoperahouse.com.},ResourceRecord {rrname = "alltcpportsopen.firewall.test.zonetransfer.me.", rrtype = A, rrclass = 1, rrttl = 301, rdata = 127.0.0.1},ResourceRecord {rrname = "testing.zonetransfer.me.", rrtype = CNAME, rrclass = 1, rrttl = 301, rdata = www.zonetransfer.me.},ResourceRecord {rrname = "vpn.zonetransfer.me.", rrtype = A, rrclass = 1, rrttl = 4000, rdata = 174.36.59.154},ResourceRecord {rrname = "www.zonetransfer.me.", rrtype = A, rrclass = 1, rrttl = 7200, rdata = 5.196.105.14},ResourceRecord {rrname = "xss.zonetransfer.me.", rrtype = TXT, rrclass = 1, rrttl = 300, rdata = '><script>alert('Boo')</script>}]
```
vdukhovni: This example is much too wide, and hardcodes volatile data. Any test of AXFR should be in a test module, rather than be a "doctest". Ideally we could spin up a test server to talk to, so we're not relying on stability of data on the public internet.

Network/DNS/Lookup.hs
```diff
@@ -429,3 +430,26 @@ lookupSRV rlv dom = do
...
+
+lookupAXFR :: Resolver -> Domain -> IO (Either DNSError [ResourceRecord])
+lookupAXFR rlv dom = do
+  erds <- DNS.lookupRaw rlv dom AXFR
```
vdukhovni: I think that `AXFR` should use its own low-level "lookup" interface, rather than lookupRaw, and should probably return a stream, rather than a list of records.  The zones I AXFR have millions of records, and should not be stored as a single array in memory.

Network/DNS/Transport.hs
```diff
@@ -118,11 +117,15 @@ resolveOne ai gen q tm retry ctls rcv =
...
+--
 udpTcpLookup :: IO Identifier -> UdpRslv
 udpTcpLookup gen retry rcv ai q tm ctls = do
     ident <- gen
-    udpLookup ident retry rcv ai q tm ctls `E.catch`
-            \TCPFallback -> tcpLookup gen ai q tm ctls
+    case qtype q of
+      AXFR -> tcpLookup gen ai q tm ctls
+      _    -> udpLookup ident retry rcv ai q tm ctls `E.catch`
+                 \TCPFallback -> tcpLookup gen ai q tm ctls
```
vdukhovni: Write a new low-level interface for AXFR, rather than changing the "lookup" interface.  AXFR is not a "lookup" for a single RRset.

Network/DNS/Transport.hs
```diff
@@ -173,7 +176,7 @@ udpLookup ident retry rcv ai q tm ctls = do
     --
     getAns sock = do
         mres <- rcv sock
-        if checkResp q ident mres
+        if checkResp ident mres
```
vdukhovni: This change must go.

Network/DNS/Transport.hs
```diff
@@ -202,18 +205,39 @@ tcpLookup gen ai q tm ctls =
         else return res
   where
     addr = addrAddress ai
+    receiveAndCheck vc ident = do
+        resp <- receiveVC vc
+        if checkResp ident resp
+        then return resp
+        else E.throwIO SequenceNumberMismatch
```
vdukhovni: This change must go.

Network/DNS/Transport.hs
```diff
@@ -202,18 +205,39 @@ tcpLookup gen ai q tm ctls =
...
+                    -- first DNS message contains a single answer, which has to
+                    -- be the SOA record.
+                    xs <- if length (answer x) > 1 && isAXFRLast x
+                          then return []
+                          else repeatUntilM isAXFRLast (receiveAndCheck vc ident)
+                    -- We remove the last answer record which is a repeat of the
+                    -- SOA record, also sent as the first answer record.
+                    return $ x { answer = init $ concatMap answer (x:xs) }
+                _ -> receiveAndCheck vc ident
+        maybe (E.throwIO TimeoutExpired) return mres
```
vdukhovni: Please move all this logic into a separate "lookupAXFR" interface.

Network/DNS/Transport.hs
```diff
@@ -232,3 +256,15 @@ isIllegal dom
...
+repeatUntilM predicate x = go []
+  where go acc = do
+          val <- x
+          let acc' = val:acc
+          if predicate val
+            then return $ reverse acc'
+            else go acc'
+
+isAXFRLast :: DNSMessage -> Bool
+isAXFRLast = (== SOA) . rrtype . last . answer
```
vdukhovni: AXFR should present a monadic data stream, not a pure list.  If the caller then chooses to buffer the data in memory, that's the caller's choice, but the API should present a constant-space interface.

Network/DNS/Types.hs
```diff
@@ -244,10 +245,13 @@ pattern CDNSKEY    = TYPE  60 -- RFC 7344
...
 pattern CSYNC :: TYPE
 pattern CSYNC      = TYPE  62 -- RFC 7477
+-- | Zone transfer (RFC5936)
+pattern AXFR :: TYPE
+pattern AXFR       = TYPE 252 -- RFC 5936
 -- | A request for all records the server/cache has available
 pattern ANY :: TYPE
 pattern ANY        = TYPE 255
--- | A request for all records the server/cache has available
+-- | Certificate Authority Authorization
```
vdukhovni: Please add an RFC reference <https://tools.ietf.org/html/rfc6844> and use the correct name: "Certification Authority Authorization".

Network/DNS/Transport.hs
```diff
@@ -19,12 +19,11 @@ import Network.DNS.Imports
...
 
--- | Check response for a matching identifier and question.  If we ever do
+-- | Check response for a matching identifier.  If we ever do
 -- pipelined TCP, we'll need to handle out of order responses.  See:
 -- https://tools.ietf.org/html/rfc7766#section-7
-checkResp :: Question -> Identifier -> DNSMessage -> Bool
-checkResp q seqno resp =
-   identifier (header resp) == seqno && [q] == question resp
+checkResp :: Identifier -> DNSMessage -> Bool
+checkResp seqno resp = identifier (header resp) == seqno
```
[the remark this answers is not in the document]
dudebout: Agreed.

Network/DNS/Lookup.hs
dudebout: Agreed. I did not intend to get it on master, but wanted to have something to start the discussion.

Network/DNS/Lookup.hs
dudebout: Makes sense

Network/DNS/Types.hs
dudebout: Will do

dudebout:
Thanks for the detailed feedback @vdukhovni. I'll work on improvements.

For the iterator improvements, did you have something like `conduit` and `pipes` in mind, or something different?

kazu-yamamoto:
If possible, `dns-conduit` or `dns-pipe` should be implemented atop `dns`. 
`dns` should provide low APIs which is sufficient for them.

vdukhovni:
Yes, conduit is what I had in mind, but did not state. I use conduits in all my code.  However, Network.DNS should not have a Conduit dependency, rather a package that provides a conduit for AXFR should be separate from Network.DNS and depend on that and Conduit.  That new package (as @kazu-yamamoto suggested) can use the low-level iterator for AXFR results provided by Network.DNS.

dudebout:
@vdukhovni I took a shot at streaming AXFR in https://github.com/dudebout/dns/commit/eb6dc571b42e64f23bd53be7c96b824a900be171. Could you take a look and let me know if it looks like what you had in mind? If so, I'll work on high level documentation and cleanups.

Notes:
- I have added a Conduit module using the interface and an executable using that interface for demonstration purposes only. It will not be in the final PR.
- As I was working on it, I found it easier to use exceptions rather than weave `Either DNSError` around (it also follows the advice from https://www.fpcomplete.com/blog/2016/11/exceptions-best-practices-haskell to not use `EitherT error` around IO). What is the rationale for having `Either`s? Do we want `Either`s for the streaming? If so, I'll work on the change.

dudebout:
@vdukhovni @kazu-yamamoto , I took a stab at a streaming AXFR implementation in https://github.com/dudebout/dns/commit/eb6dc571b42e64f23bd53be7c96b824a900be171. Does that look like what you guys had in mind?

kazu-yamamoto:
@vdukhovni May I ask to review this?

kazu-yamamoto:
Gentle ping @vdukhovni

vdukhovni:
Closed. A fresh start is needed.

vdukhovni:
Please open a new pull request based on eb6dc571b42e64f23bd53be7c96b824a900be171

PR #135, opened 2019-02-25
add AXFR types and error cases

dudebout:
@vdukhovni I have implemented some of your suggested changes from #134 in this new PR:

- documented CAA, including mention of the RFC
- added AXFR types
- added an error constructor to prevent sending an AXFR request from the existing interface
- added an error constructor to differentiate between sequence number and question mismatch
- made the question validation optional in the new `checkRespM`, for future zone transfer use

### DIFF
--- a/Network/DNS/Transport.hs
+++ b/Network/DNS/Transport.hs
@@ -23,17 +23,13 @@ import Network.DNS.Types.Internal
 -- pipelined TCP, we'll need to handle out of order responses.  See:
 -- https://tools.ietf.org/html/rfc7766#section-7
 checkResp :: Question -> Identifier -> DNSMessage -> Bool
-checkResp q seqno = isNothing . checkRespM (Just q) seqno
+checkResp q seqno = isNothing . checkRespM q seqno
 
-checkRespM :: Maybe Question -> Identifier -> DNSMessage -> Maybe DNSError
-checkRespM qM seqno resp =
-    if identifier (header resp) /= seqno
-    then Just SequenceNumberMismatch
-    else do
-      qs <- (:[]) <$> qM
-      if qs /= question resp
-      then Just QuestionMismatch
-      else Nothing
+checkRespM :: Question -> Identifier -> DNSMessage -> Maybe DNSError
+checkRespM q seqno resp
+  | identifier (header resp) /= seqno = Just SequenceNumberMismatch
+  | [q] /= question resp              = Just QuestionMismatch
+  | otherwise                         = Nothing
 
 ----------------------------------------------------------------
 
@@ -221,7 +217,7 @@ tcpLookup gen ai q tm ctls =
             receiveVC vc
         case mres of
             Nothing  -> E.throwIO TimeoutExpired
-            Just res -> maybe (return res) E.throwIO (checkRespM (Just q) ident res)
+            Just res -> maybe (return res) E.throwIO (checkRespM q ident res)
 
 ----------------------------------------------------------------
 
--- a/Network/DNS/Transport.hs
+++ b/Network/DNS/Transport.hs
@@ -23,8 +23,17 @@ import Network.DNS.Types.Internal
 -- pipelined TCP, we'll need to handle out of order responses.  See:
 -- https://tools.ietf.org/html/rfc7766#section-7
 checkResp :: Question -> Identifier -> DNSMessage -> Bool
-checkResp q seqno resp =
-   identifier (header resp) == seqno && [q] == question resp
+checkResp q seqno = isNothing . checkRespM (Just q) seqno
+
+checkRespM :: Maybe Question -> Identifier -> DNSMessage -> Maybe DNSError
+checkRespM qM seqno resp =
+    if identifier (header resp) /= seqno
+    then Just SequenceNumberMismatch
+    else do
+      qs <- (:[]) <$> qM
+      if qs /= question resp
+      then Just QuestionMismatch
+      else Nothing
 
 ----------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


dudebout:
@vdukhovni I have implemented some of your suggested changes from #134 in this new PR:

- documented CAA, including mention of the RFC
- added AXFR types
- added an error constructor to prevent sending an AXFR request from the existing interface
- added an error constructor to differentiate between sequence number and question mismatch
- made the question validation optional in the new `checkRespM`, for future zone transfer use

Network/DNS/Transport.hs
```diff
@@ -23,8 +23,17 @@ import Network.DNS.Types.Internal
 -- pipelined TCP, we'll need to handle out of order responses.  See:
 -- https://tools.ietf.org/html/rfc7766#section-7
 checkResp :: Question -> Identifier -> DNSMessage -> Bool
-checkResp q seqno resp =
-   identifier (header resp) == seqno && [q] == question resp
+checkResp q seqno = isNothing . checkRespM (Just q) seqno
```
vdukhovni: There's no reason to complicate this interface. Some duplication of code is harmless in the name of simplicity. The function that does not check the question can repeat the code to check the transaction id.

Network/DNS/Transport.hs
```diff
@@ -23,8 +23,17 @@ import Network.DNS.Types.Internal
...
+
+checkRespM :: Maybe Question -> Identifier -> DNSMessage -> Maybe DNSError
+checkRespM qM seqno resp =
+    if identifier (header resp) /= seqno
+    then Just SequenceNumberMismatch
+    else do
+      qs <- (:[]) <$> qM
+      if qs /= question resp
+      then Just QuestionMismatch
+      else Nothing
```
vdukhovni: The new function should not take a question, no reason for `Maybe` arguments.

Network/DNS/Transport.hs
```diff
@@ -70,6 +79,7 @@ type UdpRslv = Int -- Retry
 resolve :: Domain -> TYPE -> Resolver -> Rslv0
 resolve dom typ rlv qctls rcv
   | isIllegal dom = return $ Left IllegalDomain
+  | typ == AXFR   = return $ Left NonZoneTransferAXFRRequest
```
vdukhovni: This is good. Thanks.

Network/DNS/Transport.hs
```diff
@@ -210,10 +220,8 @@ tcpLookup gen ai q tm ctls =
             sendVC vc qry
             receiveVC vc
         case mres of
-            Nothing                     -> E.throwIO TimeoutExpired
-            Just res
-                | checkResp q ident res -> return res
-                | otherwise             -> E.throwIO SequenceNumberMismatch
+            Nothing  -> E.throwIO TimeoutExpired
+            Just res -> maybe (return res) E.throwIO (checkRespM (Just q) ident res)
```
vdukhovni: There's no reason to change this code.  AXFR will use a different code path, because it needs to return a stream of resource records, with an IO interface to retrieve results that don't fit in a single 64k buffer.

Network/DNS/Transport.hs
dudebout: The main purpose of this new function is to differentiate between sequence number and question mismatch. I can make it take a `Question` instead of a `Maybe Question.

Network/DNS/Transport.hs
dudebout: This change is unrelated to AXFR (modulo the `Just` that I will remove). This change ensures that sequence number and question mismatches do not get conflated under `SequenceNumberMismatch`.

dudebout:
I have made the change related to the `Maybe Question`.

I was thinking of doing the documentation and new function in a different PR. The PR as it stands, is a self-contained unit. It documents the current state of the `dns` library with respect to AXFR. It was more of a documentation cleanup PR.

Network/DNS/Transport.hs
```diff
@@ -70,6 +75,7 @@ type UdpRslv = Int -- Retry
 resolve :: Domain -> TYPE -> Resolver -> Rslv0
 resolve dom typ rlv qctls rcv
   | isIllegal dom = return $ Left IllegalDomain
+  | typ == AXFR   = return $ Left NonZoneTransferAXFRRequest
```
vdukhovni: Bikeshedding: What do you think of `InvalidAXFRLookup` instead of the longer `NonZoneTransferAXFRRequest`? Matter of taste, the original is not confusing or anything, just a bit long...

Network/DNS/Types.hs
```diff
@@ -375,6 +383,13 @@ data DNSError =
     -- | The sequence number of the answer doesn't match our query. This
     --   could indicate foul play.
     SequenceNumberMismatch
+    -- | The question section of the response doesn't match our query. This
+    --   could indicate foul play.
+  | QuestionMismatch
+    -- | A zone tranfer, i.e., a request of type AXFR, was attempted with the
+    -- "lookup" interface. Zone transfer is different enough from "normal"
+    -- requests that it requires a different interface.
+  | NonZoneTransferAXFRRequest
```
vdukhovni: See bikeshedding comment.

Network/DNS/Transport.hs
dudebout: Makes sense to me. I made the change

kazu-yamamoto:
@dudebout Thank you for your contribution!

kazu-yamamoto:
@vdukhovni Thank you for merging. Please use `--no-ff` to create a merge commit which refers to this issue (including the word `#NNN`) next time.

vdukhovni:
@kazu-yamamoto For the record, I find "merge" commits highly objectionable. They make it very difficult to reason about the change history. All my projects use rebase only, with a linear history.  If we want the PR number in the commit, we can ask the author to add it, or we can modify the final commit as part of the rebase and add "Reviewed by:", ... comments, etc.  I'd very much like to avoid actual merges at all costs.

kazu-yamamoto:
I prefer a *liner* history *with* merge commits as we can find in `hs-tls`.

- This style nicely makes groups for related commits.
- With a good Git viewer, we can see the final form of changes on the merge commit.

Do you think it is diffcult to reason about the change history even with this style?

PR #136, opened 2019-03-21
Add rrlength to ResourceRecord.

j6carey:
This field may help identify inefficient or malicious encodings.

When decoding, rrlength tracks the length of the payload of
the resource record as specified in the resource record header,
regardless of whether that length is the minimum necessary.

Encoding ignores rrlength; when you construct a ResourceRecord
for the purpose of encoding it, you may set rrlength to any value.

### DIFF
--- a/Network/DNS/Decode/Parsers.hs
+++ b/Network/DNS/Decode/Parsers.hs
@@ -59,7 +59,7 @@ getResponse = do
         -- | Extract EDNS information from an OPT RR.
         --
         optEDNS :: ResourceRecord -> Maybe (EDNS, Word16)
-        optEDNS (ResourceRecord "." OPT udpsiz ttl' (RD_OPT opts)) =
+        optEDNS (ResourceRecord "." OPT udpsiz ttl' _ (RD_OPT opts)) =
             let hrc      = fromIntegral rc .&. 0x0f
                 erc      = shiftR (ttl' .&. 0xff000000) 20 .|. hrc
                 secok    = ttl' `testBit` 15
@@ -133,7 +133,7 @@ getResourceRecord = do
     ttl <- get32
     len <- getInt16
     dat <- fitSGet len $ getRData typ len
-    return $ ResourceRecord dom typ cls ttl dat
+    return $ ResourceRecord dom typ cls ttl (fromIntegral len) dat
 
 ----------------------------------------------------------------
 
--- a/Network/DNS/Encode/Builders.hs
+++ b/Network/DNS/Encode/Builders.hs
@@ -12,6 +12,7 @@ module Network.DNS.Encode.Builders (
   , putDomain
   , putMailbox
   , putResourceRecord
+  , putRData
   ) where
 
 import Control.Monad.State (State, modify, execState, gets)
@@ -57,8 +58,9 @@ putDNSMessage msg = putHeader hd
       where
         prependOpt ads = mapEDNS eh (fromEDNS ads $ fromRCODE rc) ads
           where
+            -- rrlength goes unused during encoding; hence we just use 0 here:
             fromEDNS :: AdditionalRecords -> Word16 -> EDNS -> AdditionalRecords
-            fromEDNS rrs rc' edns = ResourceRecord name' type' class' ttl' rdata' : rrs
+            fromEDNS rrs rc' edns = ResourceRecord name' type' class' ttl' 0 rdata' : rrs
               where
                 name'  = BS.singleton '.'
                 type'  = OPT
--- a/Network/DNS/Encode/Builders.hs
+++ b/Network/DNS/Encode/Builders.hs
@@ -12,7 +12,6 @@ module Network.DNS.Encode.Builders (
   , putDomain
   , putMailbox
   , putResourceRecord
-  , putRData
   ) where
 
 import Control.Monad.State (State, modify, execState, gets)
--- a/Network/DNS/Encode/Builders.hs
+++ b/Network/DNS/Encode/Builders.hs
@@ -57,7 +57,7 @@ putDNSMessage msg = putHeader hd
       where
         prependOpt ads = mapEDNS eh (fromEDNS ads $ fromRCODE rc) ads
           where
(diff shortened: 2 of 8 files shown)


j6carey:
No problem; thanks for your reply.

Because of domain name compression, the original value of the length field cannot be recovered: the same Haskell value can result from decoding resource records of distinct lengths.  At best one could (with quite a bit of code) infer the length that would result if compression was used to the maximum extent possible.

An inefficient or malicious sender might fail to exploit compression to the maximum extent possible, either because of software flaws or out of a desire to tunnel illicit data through DNS.  Looking at the parse result, one cannot see detect such situations.  Hence the additional field.

If you feel that this feature is too special-purpose for this library, I understand.  It would be convenient for me, but my needs are unusual.

kazu-yamamoto:
OK. I have no objection to merge this PR.

@vdukhovni What do you think?

vdukhovni:
My instinct is that this PR is not a good idea. The use-case is too narrow.

Why retain just lengths? A more broadly useful PR would (optionally) expose the `ByteString` of the complete original wire form of the DNS response and the protocol + address of the resolver that gave the answer. But it should be able to do this even (and perhaps especially) when we're returning a decoding error. But we don't return a DNSMessage when there's a decoding error.

Therefore, to support better visibility of issues with the original encoding we'd need a low-level resolver function that (optionally) returns the complete wire-form response `ByteString` payload (with TCP include the 2 leading length octets, which allows returning truncated responses) in addition to the decoded result or an error.

```haskell
data DNSResponse = DNSResponse (Either DNSError DNSMessage) (Maybe (Protocol, Address, ByteString))

lookupRawMessage :: Resolver -> Domain -> TYPE -> QueryControls -> IO DNSResponse

lookupRawCtl :: Resolver -> Domain -> TYPE -> QueryControls -> IO (Either DNSError DNSMessage)
lookupRawCtl = (((fmap getMessage .) .) .) . lookupRawMessage
  where
    getMessage (DNSResponse a _) = a
```

This has the advantage that we don't need to change the ResourceRecord structure, and return exactly the same data as before from `lookupRawCtl` and its callers. Those who want more control can call `lookupRawMessage`.

That still leaves the requestor with the task of parsing the `ByteString`, but since raw lengths are desired before decompression, this should not be difficult. Just parse the header, and then each of the sections, returning raw, undecoded rrdata buffers.

j6carey:
In my use case I already have the `ByteString` corresponding to the DNS message as a whole; I do not need a feature to provide that.  If the best answer is that I write a separate parser to get the length info from the raw DNS message, then we can just close this PR (unless you wish to make that change to accomplish some other goal).  Thanks for your consideration of this issue.

vdukhovni:
> @j6carey If you already have the message octets, and want the raw rdata lengths (no decompression, ...), that would seem to require exceedingly little parsing.  You just skip the header, `qdCount`, `anCount`, `nsCount` and `arCount` and then iterate through the questions and RRs without bothering to decompress the qname (pointer labels terminate the domain name). You then easily recover the rdata lengths.  I don't think it makes sense to change the library to embed these in the resource records returned by the primary DNSMessage decoder.

That said,  we could consider providing an alternative decoder for ByteStrings provided by the caller, which returns a raw header + lists of (optionally decompressed) qnames and uninterpreted question, answer, authority and "additional" rdata blobs.

I just don't think it makes sense to change the existing `ResourceRecord` to carry extra information almost nobody needs. There could be a separate parser returning `RawQuestion` or `RawAnswer` records 

```haskell
-- Wire size == ByteString.length(rawQname) + 4
data RawQuestion = RawQuestion
    { rawQnameLen :: !Int16
    , rawQtype    :: !Int16
    , rawQClass   :: !Int16
    }

-- Wire size == rawOwnerLen + rawRDLength + 10
data RawAnswer = RawAnswer
    { rawOwnerLen  :: !Int16
    , rawType      :: !Int16
    , rawClass     :: !Int16
    , rawTTL       :: !Int32
    , rawRDLength  :: !Int16
    , rawRData     :: !ByteString
    }

-- Uncompress domain name at given offset
getDomain :: ByteString -- ^ containing message
          -> Int16      -- ^ name offset

-- Raw decode of question (typically 12 bytes from start of message)
getRawQuestion :: ByteString -- ^ DNS message buffer
               -> Int        -- ^ Question offset in message
               -> Either DNSError RawQuestion

-- Raw decode of RR at given offset from start of message
getRawAnswer :: ByteString -- ^ DNS message buffer
             -> Int        -- ^ RR offset in message
             -> Either DNSError RawAnswer
```

We might then layer the higher level decoders over these, throwing away the intermediate raw representation, or just offer these as a parallel interface.

vdukhovni:
@j6carey Any interest in implementing something along the above lines?

vdukhovni:
All this said, what is the actual motivation for capturing the RData lengths?

j6carey:
@vdukhovni , thanks, those are some interesting ideas.

As to motivation: the length figure may prove useful in detecting inefficient or malicious encodings.

As to implementation: at minimum I could write the skeletal parser you mentioned, the one that just picks out the resource record payload lengths from the message `ByteString`.  That would probably fulfill my needs without burdening the `dns` library.

Making an alternative "raw" parser of the type you suggested is tempting, but after thinking about it a bit my intuition is that maintaining two parsers that are unconnected is a bit fragile, and implementing the full parser on top of the "raw" parser, while possible, would require benchmarks and perhaps optimization to ensure that it does not slow down typical uses of this library.  I'm afraid that at the moment I can't justify the development effort to attempt that, though I may return to the issue in future.

Thanks very much for your thoughts on this.

vdukhovni:
Just for the record, "inefficient" encodings are in the eye of the beholder. See

https://mailarchive.ietf.org/arch/msg/dnsop/setoq3BrtenzvTddtD1HduMIKmo

To avoid costly CPU cycles, many authoritative servers limit the kinds of compression they're willing to do to the most common useful cases.  E.g. pointers only point either into the query domain, or the first qname of an answer RRset, and not into domains in RDATA.  Strategies vary...

j6carey:
@vdukhovni , very interesting; thanks.

PR #137, opened 2019-07-05
POC of streaming AXFR with Conduit wrapper

dudebout:
Follow up from https://github.com/kazu-yamamoto/dns/pull/134. This is a POC only to determine if the approach matches @vdukhovni's expectations.

Notes:

* I have added a Conduit module using the interface and an executable using that interface for demonstration purposes only. It will not be in the final PR.
* As I was working on it, I found it easier to use exceptions rather than weave Either DNSError around (it also follows the advice from https://www.fpcomplete.com/blog/2016/11/exceptions-best-practices-haskell to not use EitherT error around IO). What is the rationale for having Eithers? Do we want Eithers for the streaming? If so, I'll work on the change.

### DIFF
--- a/Network/DNS/ZoneTransfer.hs
+++ b/Network/DNS/ZoneTransfer.hs
@@ -1,0 +1,87 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Network.DNS.ZoneTransfer (
+    RRStream
+  , initiateZoneTransfer
+  , nextRecord
+  , closeRRStream
+  ) where
+
+import qualified Data.ByteString.Char8 as BS
+import           Data.IORef            (IORef, newIORef, readIORef, writeIORef)
+import qualified Data.List.NonEmpty    as NE
+import           Data.Maybe            (fromMaybe)
+import           Network.Socket        (AddrInfo, Socket, addrAddress, close,
+                                        connect)
+
+import Network.DNS.IO
+import Network.DNS.Transport
+import Network.DNS.Types
+import Network.DNS.Types.Internal
+
+data RRStream = RRStream
+  { socket    :: Socket
+  , timeout   :: Int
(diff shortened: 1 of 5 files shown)


Network/DNS/ZoneTransfer.hs
```diff
@@ -0,0 +1,87 @@
...
+        nextRecord rrst
+
+closeRRStream :: RRStream -> IO ()
+closeRRStream = close . socket
+
+receiveOneMessage :: RRStream -> IO ()
+receiveOneMessage rrst = do
+  -- FIXME wrap timeout
+  msg <- receiveVC $ socket rrst
+  let isLast = (rrtype . last . answer) msg == SOA
```
vdukhovni: This code needs to deal with error responses, and must validate the first answer's question section (https://tools.ietf.org/html/rfc5936#section-2.2) and id. The RFC also requires a matching question section in explicit error responses.

If TSIG is implemented, then response TSIGs need to be checked, that's likely _phase 2_.

dudebout:
Thank you Viktor for the detailed feedback. I will reply to it and start implementing some of the suggestions in the next few weeks.

vdukhovni:
@dudebout, are you still interested in getting this done, or should I take over at some point?

dudebout:
@vdukhovni I am still interesting in getting this done but have not found the time yet. If you have a time constraint, let me know.

kazu-yamamoto:
No time constraint at least to me!

vdukhovni:
> No time constraint at least to me!

Same here, just did not want to simply ignore this forever. You can take some more time, but I may beat you to it some time in Q1/Q2 2020 if you don't get to it by then.

dudebout:
Sounds good. I'll let you know when I start working on it.

PR #145, opened 2019-10-16
Work around stack sandbox differences

vdukhovni:
Stack sanboxes and Cabal sanboxes behave differently for internal packages.  In the case of stack the internal package is hidden by default, and the package-id values are very different.

So to make the "cbits" visible to doctests on Windows we needs to link with just the main library, not the internal library which is rather difficult to specify portably between cabal and stack.

Not having access to the internal library has some drawbacks, We need to expose the internal library to the doctests in source form, which result in those tests running twice when doctest looks for modules to test, it finds the internal modules both directly and indirectly...

Don't know of a better solution for now.  It would be great if stack and cabal were better aligned here.


kazu-yamamoto:
@snoyberg Is this well-known issue? Could you suggest any workaround?

vdukhovni:
> @snoyberg Is this well-known issue? Could you suggest any workaround?

I did find a work-around, by asking for the doctest fragments to be linked only to the exposed `dns` library package (had to switch from `package-id` package names, as the former differ between `stack` and `cabal`), and not also asking for the `dns-internal` library.

It seems that stack calls both the exposed and internal libraries "dns-4.0.0", but in different package databases, and with different package ids (that instead of cabal's `inplace` tag, include long hashed tags), and when stack links executables it passes the appropriate `-package-id` arguments. Stack keeps the internal package hidden by default.

Cabal also optionally populates a .ghc.environment.* file, that makes it easy to find the relevant packages, so I was able to construct a more precise set of arguments for linking doctests, that also managed to avoid testing some of the internal modules twice.  I found the new cabal machinery somewhat easier to work with (hook into).

Switching to the portable (just the package name) variant at the moment runs some doc tests more than once, but that's pretty harmless.  The below manages to link the doctests even on Windows where the `dns` package is linked with the `dns-internal` package, which in turn contains the necessary `cbits` code to make DNS be able to find the default nameservers on windows. Not having the internal package forces me to add `-iInternal` to the options so that `ghci` can find the internal dependencies of the souces it is interpreting, which then causes the double testing of the internal modules.  With just the original cabal-specific approach I was able to avoid test duplication.

What I'm testing now is:
```haskell
-- | Run doctests only on non-Windows systems with GHC 8.4 or later.
--      
-- The Windows doctests now compile and run, but either succeed quickly or
-- randomly hang (before AppVeyor kills them after an hour).  So we disable
-- these for now.  This can be tested again at some point in the future.
--          
module Main where
        
#if !defined(mingw32_HOST_OS) && MIN_TOOL_VERSION_ghc(8,4,0)
import Test.DocTest
import System.Environment
            
-- | Expose precompiled library modules.
modules :: [String]
modules =   
  [ "-XOverloadedStrings"
  , "-XCPP" 
  , "-i","-isrc"
  , "-O", "-threaded"
  , "-package=dns"
  , "src/Network/DNS.hs"
  ]                 
            
main :: IO ()
main = getArgs >>= doctest . (++ modules)
            
#else

main :: IO ()
main = return ()

#endif
```

snoyberg:
I know basically nothing of internal libraries and how they're implemented in Stack, sorry.

vdukhovni:
> I know basically nothing of internal libraries and how they're implemented in Stack, sorry.

Thanks for answering promptly. FWIW, I like some aspects of the cabal approach, it looks more easily adaptable.

That said, cabal has no notion of LTS snapshots, without which assembling a coherent collection of dependencies can be a pain. So many thanks for your contributions to stack, much appreciated, even if we complain now and then!

vdukhovni:
I think this is ready now.

The final step was getting haddock to not leave out the documentation from `Network.DNS.Types.Internal`. We need this module for all the other internal modules, but it also provides the types for the public user interface, so its documentation needs to be visible. 

At this point we have working tests with both cabal and stack, and now also GHC 8.8.1 on Linux! Also (non-doctest) tests pass on Windows, and even the doctests always compile and occasionally succeed.  Should we be testing more GHC versions on Windows?  Are more such versions available?

So if you don't see any issues you'd like addressed (perhaps even cosmetic choices like the directory names), this can perhaps be merged. Otherwise, we could still move things around a bit, perhaps:

* public modules under `src/`
* internal modules under `src/internal`

Would that be any better (I suspect the extra parent directory buys us little).

kazu-yamamoto:
`cabal v1-test` fails on my local Mac:

```
Test suite doctests: RUNNING...
doctests: <command line>: cannot satisfy -package dns
    (use -v for more information)
Test suite doctests: FAIL
Test suite logged to: dist/test/dns-4.0.0-doctests.log
2 of 3 test suites (2 of 3 test cases) passed.
cab: callCommand: cabal v1-test (exit 1): failed
```

vdukhovni:
What version of GHC and cabal do you have on the Mac?  The package should be found with GHC 8.4 and later, because cabal should be creating the `.ghc.environment.*` file that sets up the required package databases. If `v1-test` does not work, perhaps we can just not support it?

I'd like to keep open the possibility of having the doctests ultimately work on Windows, where `-package dns` is needed to load the "cbits" code.

kazu-yamamoto:
Both GHC 8.6.5 and GHC 8.8.1 fail.

vdukhovni:
I looked further into the `v1-test` issue, and it works if I run the executable with an extra `-package-db=dist/package.conf.inplace` argument. With `v1-build` and `v1-test` there is no `.ghc.environment` file with the location of the package database...

My preference is to not support the `v1` interface, but if you think that's important then we have to unconditionally forgo doctests on Windows.  Supporting 3 different build frameworks + doctest (which is in a sense yet another) is non-trivial...

vdukhovni:
@kazu-yamamoto How shall we proceed? We have to do something, since without this PR we not only break `cabal v1-test` but also `stack test`. Our choice is either forget about doctests on Windows, or accept `cabal v1-test` breakage, unless there's some clever way of conditionally doing something special for `cabal v1`.

PR #146, opened 2019-10-20
Test without internal library + -package=dns for Windows doctest

vdukhovni:
This configuration passes "cabal v1-test", "cabal v2-test" and "stack test", and *builds* and *runs* doctests on Windows, which then fail, but at least the "cbits" code is now loaded, so it could start working later, when the source of the hang is resolved.

Sadly, it precludes using  "internal" libraries, which are not compatible with a Custom build at present.

Which do we want more?

* An internal library, or
* "cabal v1-test" working

By using "cabal-doctest" all the required package-db, ... flags are passed in, making "-package=dns" work.


kazu-yamamoto:
@vdukhovni Is this unnecessary anymore since #145 is merged?

vdukhovni:
> @vdukhovni Is this unnecessary anymore since #145 is merged?

No, not _necessary_. This is a  possible alternative, which fixes `v1-test` at the cost of not having an internal library. So it is either this, or #145, not both.

kazu-yamamoto:
OK. Let's close this.

PR #147, opened 2019-11-13
RFC-1996 NOTIFY DNS Message template

iamemhn:
Create a NOTIFY DNS Message for a particular zone, following RFC-1996.

### DIFF
--- a/internal/Network/DNS/Types/Internal.hs
+++ b/internal/Network/DNS/Types/Internal.hs
@@ -1330,6 +1330,35 @@ defaultResponse = DNSMessage {
   , additional = []
   }
 
+-- | A 'DNSMessage' template for NOTIFY queries for the given zone,
+-- as per RFC-1996.
+
+notifyMessage :: ByteString -> DNSMessage
+notifyMessage zoneName = DNSMessage {
+  header = DNSHeader {
+      identifier = 0
+    , flags = DNSFlags {
+             qOrR = QR_Query
+           , opcode = OP_NOTIFY
+           , authAnswer = True
+           , trunCation = False
+           , recDesired = True
+           , recAvailable = False
+           , rcode = NoErr
+           , authenData = False
+           , chkDisable = False
+      }
+    }
+    , ednsHeader = EDNSheader defaultEDNS
+    , question = [ Question { qname = zoneName
+                            , qtype = SOA
+                            }
+                 ]
+    , answer = []
+    , authority = []
+    , additional = []
+  }
+
 -- | A query template with 'QueryControls' overrides applied,
 -- with just the 'Question' and query 'Identifier' remaining
 -- to be filled in.
--- a/internal/Network/DNS/Types/Internal.hs
+++ b/internal/Network/DNS/Types/Internal.hs
@@ -1330,28 +1330,28 @@ defaultResponse = DNSMessage {
   , additional = []
   }
 
--- | A 'DNSMessage' template for NOTIFY queries for the given zone,
--- as per RFC-1996.
+-- | A 'DNSMessage' template for NOTIFY Request queries for the
+-- given zone, as per RFC-1996 (Section 4.5).
 
-notifyMessage :: ByteString -> DNSMessage
-notifyMessage zoneName = DNSMessage {
+notifyRequestMessage :: ByteString -> DNSMessage
+notifyRequestMessage zoneName = DNSMessage {
   header = DNSHeader {
       identifier = 0
     , flags = DNSFlags {
              qOrR = QR_Query
-           , opcode = OP_NOTIFY
-           , authAnswer = True
-           , trunCation = False
-           , recDesired = True
-           , recAvailable = False
-           , rcode = NoErr
-           , authenData = False
-           , chkDisable = False
+           , opcode = OP_NOTIFY                -- RFC-1996 4.5
+           , authAnswer = True                 -- RFC-1996 4.5
+           , trunCation = False                -- RFC-1996 3.2
+           , recDesired = False                -- RFC-1996 3.2
+           , recAvailable = False              -- RFC-1996 3.2
+           , rcode = NoErr                     -- RFC-1996 4.5
+           , authenData = False                -- RFC-1996 3.2
+           , chkDisable = False                -- RFC-1996 3.2
       }
     }
-    , ednsHeader = EDNSheader defaultEDNS
-    , question = [ Question { qname = zoneName
-                            , qtype = SOA
+    , ednsHeader = NoEDNS
+    , question = [ Question { qname = zoneName -- RFC-1996 4.5
+                            , qtype = SOA      -- RFC-1996 4.5
                             }
                  ]
     , answer = []
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -53,7 +53,7 @@ module Network.DNS.Types (
   , makeEmptyQuery
   , defaultQuery
   -- ** NOTIFY
-  , notifyMessage
+  , notifyRequestMessage
   -- ** Query Controls
   , QueryControls
   , rdFlag
--- a/Network/DNS/Types.hs
+++ b/Network/DNS/Types.hs
@@ -52,6 +52,8 @@ module Network.DNS.Types (
   , makeQuery
   , makeEmptyQuery
   , defaultQuery
+  -- ** NOTIFY
+  , notifyMessage
   -- ** Query Controls
   , QueryControls
   , rdFlag


kazu-yamamoto:
LGTM. But I want to wait for @vdukhovni's review.

internal/Network/DNS/Types/Internal.hs
```diff
@@ -1330,6 +1330,35 @@ defaultResponse = DNSMessage {
...
+           , authenData = False
+           , chkDisable = False
+      }
+    }
+    , ednsHeader = EDNSheader defaultEDNS
+    , question = [ Question { qname = zoneName
+                            , qtype = SOA
+                            }
+                 ]
+    , answer = []
```
vdukhovni: There should perhaps be an optional (normally SOA) ResourceRecord to all to the answer section. This helps the receiving system avoid unnecessary queries to its master if it is already at or ahead of the serial number in the provided SOA. NOTIFY without SOA is less efficient.

This suggests a signature of `notifyMessage :: ByteString -> [ResourceRecord] -> DNSMessage` or `notifyMessage :: ByteString -> DNSMessage -> Maybe ResourceRecord -> DNSMessage`. Probably the former for full generality. Callers that want to skip the SOA can just pass `[]` (else `Nothing`).

Alternatively, the second argument could be a `Maybe RData`, with the `ResourceRecord` constructed from the given zone, class `IN` and the TTL from the SOA. The result would be a `Maybe DNSMessage` with `Nothing` returned when the `RData` provided is not an `RD_SOA`.

Or some variation on the above.

vdukhovni:
It may be worth mentioning that some servers may limit NOTIFY, AXFR, ... to TSIG-signed requests/replies, but we don't have TSIG support at this time. If NOTIFY is to be fully supported, it may make sense to implement TSIG support at some point.

iamemhn:
On Wed, 13 Nov 2019 18:08:01 -0800
Viktor Dukhovni <notifications@github.com> wrote:

> vdukhovni requested changes on this pull request.
> 
> Some changes suggested.
[...]
> Why is `recDesired` set to `True`? I don't believe this matches
> RFC1996.
[...]
> +    , ednsHeader = EDNSheader defaultEDNS
> 
> I'm not sure that EDNS(0) is useful or appropriate with NOTIFY.

Thank you for pointing out those mistakes. The first has no effect, but
it's not RFC-complaint. The second is more subtle.

I have applied both changes and also renamed the function to
`notifyRequestMessage` because we might want to add
`notifyResponseMessage` as well (I'll think about that later -- don't
need that functionality just yet).

> There should perhaps be an optional (normally SOA) ResourceRecord to
> all to the answer section. This helps the receiving system avoid
> unnecessary queries to its master if it is already at or ahead of the
> serial number in the provided SOA. NOTIFY without SOA is less
> efficient.

I disagree on this one, based on my understanding of RFC-1996 3.7 and
3.11 together

   3.7. (...) If ANCOUNT>0, then the answer section represents an
   unsecure hint at the new RRset for this <QNAME,QCLASS,QTYPE>.  A
   slave receiving such a hint is free to treat equivilence of this
   answer section with its local data as a "no further work needs to be
   done" indication.  If ANCOUNT=0, or ANCOUNT>0 and the answer section
   differs from the slave's local data, then the slave should query its
   known masters to retrieve the new data.

   3.11. The only defined NOTIFY event at this time is that the SOA RR
   has changed.  Upon completion of a NOTIFY transaction for QTYPE=SOA,
   the slave should behave as though the zone given in the QNAME had
   reached its REFRESH interval (see [RFC1035]), i.e., it should query
   its masters for the SOA of the zone given in the NOTIFY QNAME, and
   check the answer to see if the SOA SERIAL has been incremented since
   the last time the zone was fetched.

is that adding the SOA is an option (as you mention), but
implementations should treat them as a hint and SHOULD query the
master's SOA *anyway*. I agree that *some* implementations could take
the SOA and compare it locally, but *most* implementations are going to
behave as 3.11. BIND9 slaves will ignore whatever SOA RR is included
with the message and query the masters; my guess is most others do. Do
you know of any implementation that actually uses that suggested SOA to
compare locally?
-- 
Ernesto Hernández-Novich - @iamemhn - Unix: Live free or die!
Geek by nature, Linux by choice, Debian of course.
If you can't aptitude it, it isn't useful or doesn't exist.
GPG Key Fingerprint = 0064 ADF5 EB5C DE16 99C1 6C56 F2A3 86B5 A757 E5A1

vdukhovni:
> I agree that *some* implementations could take the SOA and compare it locally, but *most* implementations are going to behave as 3.11. BIND9 slaves will ignore whatever SOA RR is included with the message and query the masters; my guess is most others do. Do you know of any implementation that actually uses that suggested SOA to compare locally?

Actually, BIND9 both sends the SOA RR in NOTIFY requests by default, _and_ does look at the SOA hint in received NOTIFY requests:
- Sending:
   https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/dns/zone.c#L13860
   https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/dns/zone.c#L13827
- Receiving:
  https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/dns/zone.c#L13922
  https://gitlab.isc.org/isc-projects/bind9/blob/master/lib/dns/zone.c#L14013

internal/Network/DNS/Types/Internal.hs
```diff
@@ -1330,28 +1330,28 @@ defaultResponse = DNSMessage {
   , additional = []
   }
 
--- | A 'DNSMessage' template for NOTIFY queries for the given zone,
--- as per RFC-1996.
+-- | A 'DNSMessage' template for NOTIFY Request queries for the
```
vdukhovni: Cosmetic, but `queries` is pretty much redundant following `Request`, so I'd just write `Requests` and drop the word `queries`.

vdukhovni:
I'm also curious what the use-case is for sending NOTIFY in Haskell? The `Network.DNS` library does not presently implement much of what one might need for a DNS responder (name server), and NOTIFY requests are generally sent by authoritative servers to their peers, but it seems unlikely that you have an authoritative server based on `Network.DNS`. So the need for NOTIFY support is unexpected.

[ And if the BIND9 comment above is not sufficiently clear, yes implementations are reasonably expected to skip checking the upstream SOA when a local check shows that the SOA hint in the NOTIFY does not warrant a refresh. So the BIND9 maintainers and I read the RFC text differently that you did. ]

iamemhn:
On Wed, 13 Nov 2019 22:21:57 -0800
Viktor Dukhovni <notifications@github.com> wrote:

> I'm also curious what the use-case is for sending NOTIFY in Haskell?
> The `Network.DNS` library does not presently implement much of what
> one might need for a DNS responder (name server), and NOTIFY requests
> are generally sent by authoritative servers to their peers, but it
> seems unlikely that you have an authoritative server based on
> `Network.DNS`. So the need for NOTIFY support is unexpected.

I have a tool for testing and auditing our DNS platform, that sends
artificial NOTIFY Requests to actual name servers, and also to a
companion tool that reacts to both these artificial NOTIFY Requests as
well as those sent from actual name servers. Both the trigger tool, and
the server reacting to NOTIFYs, are written in Haskell. Can't go into
details until the project completes internally.
-- 
Ernesto Hernández-Novich - @iamemhn - Unix: Live free or die!
Geek by nature, Linux by choice, Debian of course.
If you can't aptitude it, it isn't useful or doesn't exist.
GPG Key Fingerprint = 0064 ADF5 EB5C DE16 99C1 6C56 F2A3 86B5 A757 E5A1

iamemhn:
On Wed, 13 Nov 2019 18:08:01 -0800
Viktor Dukhovni <notifications@github.com> wrote:
[...] 
> There should perhaps be an optional (normally SOA) ResourceRecord to
> all to the answer section. This helps the receiving system avoid
> unnecessary queries to its master if it is already at or ahead of the
> serial number in the provided SOA. NOTIFY without SOA is less
> efficient.

On Wed, 13 Nov 2019 22:21:57 -0800
Viktor Dukhovni <notifications@github.com> wrote:
[...]
> [ And if the BIND9 comment above is not sufficiently clear, yes
> implementations are reasonably expected to skip checking the upstream
> SOA when a local check shows that the SOA hint in the NOTIFY does not
> warrant a refresh. So the BIND9 maintainers and I read the RFC text
> differently that you did. ]

I understand it better now, thank you. 

> Alternatively, the second argument could be a `Maybe RData`, with the
> `ResourceRecord` constructed from the given zone, class `IN` and the
> TTL from the SOA. The result would be a `Maybe DNSMessage` with
> `Nothing` returned when the `RData` provided is not an `RD_SOA`.

Since the NOTIFY is for one zone, I think `Maybe` fits better than a
list: either provide the SOA for the zone, or not. I'll work on that.
-- 
Ernesto Hernández-Novich - @iamemhn - Unix: Live free or die!
Geek by nature, Linux by choice, Debian of course.
If you can't aptitude it, it isn't useful or doesn't exist.
GPG Key Fingerprint = 0064 ADF5 EB5C DE16 99C1 6C56 F2A3 86B5 A757 E5A1

vdukhovni:
> Since the NOTIFY is for one zone, I think `Maybe` fits better than a list: either provide the SOA for the zone, or not. I'll work on that.

Sure.  Since the answer RRs in a NOTIFY MUST NOT be cached, the TTL on the SOA can be `0`, rather than taken from the SOA record `minimum` TTL.

A cleaner interface (avoiding a "partial" Maybe in the result type) could be achieved by taking only the serial number and filing in placeholder values for the rest of the SOA record:

You would then just combine the zone name, class `IN`, TTL `0` and the serial number input to produce the optional answer record. You could instead take a `Maybe (Domain, Mailbox, Word32, Word32, Word32, Word32, Word32)` allowing all the fields of the SOA to be specified, without a way to specify some other invalid in this context RData.
```haskell
notifyRequestMessage :: ByteString -> Maybe Word32 -> Maybe DNSMessage
notifyRequestMessage zoneName mSerial =
    message <$> answers
  where
    message answers = DNSMessage { ... }
    answers = case mSerial of
      Just serial -> [ResourceRecord { rrname = zone
                                     , rrtype = SOA
                                     , rrclass = classIN
                                     , rrttl = 0
                                     , rdata = RD_SOA "." "." serial 0 0 0 0 }]
      Nothing -> []
```

vdukhovni:
One way to make the interface still friendly, but more general, is to provide two functions. One that constructs the `DNSMessage` from the zoneName and `[ResourceRecord]` RRset, the second to construct the RRset from something resembling an SOA record (and the zoneName). There could then be more than one way to construct the answer RRset, and callers that don't have a serial number to send can just pass `[]`.

I think on balance providing two or more separate functions makes for a cleaner design:

```haskell
notifyRequestMessage :: Domain -> [ResourceRecord] -> DNSMessage
notifyRequestMessage zoneName answers = ...

notifySOASerial :: Domain -> Word32 -> ResourceRecord
notifySOASerial zoneName serial = ResourceRecord { ... }

notifySOA :: Domain -> (Domain, Mailbox, Word32, ..., Word32) -> ResourceRecord
notifySOA zoneName (mname, mrname, serial, ..., minimum) = ResourceRecord { ... }
```

vdukhovni:
One crucial question I neglected to ask is:
* Why does this function need to be in the library at all?

The `DNSMessage` type is fully exposed, and any application can construct this "NOTIFY" message directly:
```haskell
λ> :set -XOverloadedStrings
λ> import Network.DNS
λ> let x = DNSMessage {
  header = DNSHeader {
    identifier = 0,
    flags = defaultDNSFlags {
      opcode = OP_NOTIFY,
      authAnswer = True,
      recDesired = False
    }
  },
  ednsHeader = NoEDNS,
  question = [Question { qname = "example.com.", qtype = SOA} ],
  answer = [ResourceRecord {
    rrname = "example.com",
    rrtype = SOA,
    rrttl = 0,
    rrclass = classIN,
    rdata = RD_SOA "ns1.example.com." "hostmaster@example.com." 1 0 0 0 0
  }],
  authority = [],
  additional = []
}
λ> (\ (RD_SOA _ _ serial _ _ _ _) -> serial) . rdata <$> answer x
[1]
```

iamemhn:
On Sun, 17 Nov 2019 17:09:08 -0800
Viktor Dukhovni <notifications@github.com> wrote:

> One crucial question I neglected to ask is:
> * Why does this function need to be in the library at all?

Just by reading this thread, it would help library users avoid the
mistake I made with my initial implementation, which is now better
thanks to your feedback.

> The `DNSMessage` type is fully exposed, and any application can
> construct this "NOTIFY" message directly

Yet `makeQuery` and `makeEmptyQuery` justifiably exist. Presumably out
of convenience. Same can be argued about a function to easily create
*correct* NOTIFY queries. NOTIFY queries aren't created often, but
those who need them will certainly appreciate the convenience.

That was my main motivation for contributing back. Initially I send it
via private e-mail to Yamamoto-san on 2019-11-05, but then decided to
follow the formal pull-request path.
-- 
Ernesto Hernández-Novich - @iamemhn - Unix: Live free or die!
Geek by nature, Linux by choice, Debian of course.
If you can't aptitude it, it isn't useful or doesn't exist.
GPG Key Fingerprint = 0064 ADF5 EB5C DE16 99C1 6C56 F2A3 86B5 A757 E5A1

vdukhovni:
I'm inclined to not merge this code for now.  I hope you don't mind too much.

It could some day be added to a "dns-extra" package, that contains optional additional features, but as it stands, it is perhaps a bit out of place in the core library.

Putting too many ad-hoc features in the library makes it more difficult to maintain and to keep a clear focus.
But if Yamamoto-san would like to see this merged, I'm OK with that, it should at least now be reasonably correct.

kazu-yamamoto:
I would like to support the approach of the `dns-extra` package. So, let's close.
Recently I miss many important e-mail messages due to spams. 
I'm very sorry.
Github works much better for me.

Issue #150, opened 2019-11-29
dns-4.0.1 test suite does not compile


peti:
Citing from https://hydra.nixos.org/build/107456929:

~~~
Building test suite 'network-tests' for dns-4.0.1..
[1 of 3] Compiling IOSpec           ( test2/IOSpec.hs, dist/build/network-tests/network-tests-tmp/IOSpec.o )

test2/IOSpec.hs:20:9: error:
    Ambiguous occurrence ‘send’
    It could refer to either ‘Network.Socket.send’,
                             imported from ‘Network.Socket’ at test2/IOSpec.hs:5:1-21
                          or ‘DNS.send’,
                             imported from ‘Network.DNS.IO’ at test2/IOSpec.hs:8:1-28
   |
20 |         send sock qry
   |         ^^^^
~~~

felixonmars:
We have the same issue in Arch. Currently patched with the change in #152

kazu-yamamoto:
I guess that you are using old `network`. Which version? Why do you still use the old version?

felixonmars:
We are using network-2.8.0.1 at Arch. Some software are still not yet ported for network-3.

kazu-yamamoto:
OK. Your PR is a very simple. Let's merge.

PR #152, opened 2020-03-01
Fix test build failures in #150

felixonmars:


kazu-yamamoto:
I think that we don't have to support very old `network`. Should we add version-boundary to `dns.cabal`?

kazu-yamamoto:
Merged. Thank you for your contribution.
If you want to get a new version released, please let me know.

felixonmars:
It's fine for now since we already applied the patch, thank you!

PR #154, opened 2020-10-30
Avoid decode cache store/lookup for mailbox names

vdukhovni:
On rare occasions an SOA rname will match one of the domain names
in the rest of the message, with pointer compression used to
deduplicate the bytes.  But we don't use the same presentation form
for domains and mailboxes, so must not use the same cache during
message decoding.

Since SOA rnames are rarely equal to another domain name in the
message, and the SOA RR is not even present in most successful
responses, the simplest solution is to neither store nor query the
cache for mailbox names, but instead always decode them from
scratch.


kazu-yamamoto (approved):
Good catch!

kazu-yamamoto:
Rebased and merged.

PR #155, opened 2020-10-31
New splitDomain and splitMailbox functions

vdukhovni:
- Change parseLabel to return Either DNSEror ... rather
  than throw an exception
- Use the internal parseLabel in new splitDomain and splitMailbox
  functions.


kazu-yamamoto (approved):
LGTM.

kazu-yamamoto:
Merged.

PR #157, opened 2020-11-02
Add 8.10 to Travis CI matrix and update 8.8 to .4

vdukhovni:
Actually, more than the above, I want to see whether the travis CI is running at all?
It seems that we're only running AppVeyor.  Not clear why...


vdukhovni:
@kazu-yamamoto Any idea why the Travis CI is not happening?

PR #163, opened 2021-02-26
Add CAA RData decoding

kfigiela:


kazu-yamamoto:
@vdukhovni Would you review this PR?

internal/Network/DNS/Types/Internal.hs
```diff
@@ -1074,6 +1077,9 @@ instance Show RData where
       showTLSA usage selector mtype digest =
           show usage ++ " " ++ show selector ++ " " ++
           show mtype ++ " " ++ _b16encode digest
+      showCAA flags tag value =
+          show flags ++ " " ++ show tag ++ " " ++
```
vdukhovni, line 1081: The `tag` is required to be alphanumeric ASCII, and its presentation form does not employ double-quotes.  Using `show` here, will emit the `tag` in double quotes.  Unless the tag does not satisfy the required syntax, it should probably be presented without quotes.

internal/Network/DNS/Types/Internal.hs
```diff
@@ -998,6 +998,8 @@ data RData = RD_A IPv4           -- ^ IPv4 address
                                  -- ^ Child DS (RFC7344)
            | RD_CDNSKEY Word16 Word8 Word8 ByteString
                                  -- ^ Child DNSKEY (RFC7344)
+           | RD_CAA Word8 ByteString ByteString
```
vdukhovni, line 1001: Show the `tag` field be a `case-insenstive` ByteString (from `Data.CaseInsensitive`)? Or is case-insensitive matching best left to users?

vdukhovni (commented):
No major issues, but see comments.

vdukhovni:
@kfigiela any chance you'll take a look at the review comments?

kfigiela:
@vdukhovni I will try to address this next week when I'm back from vacation.

kfigiela:
@vdukhovni @kazu-yamamoto I've applied the review comments as well as rebased on top of the current master.

vdukhovni (approved):
LGTM

internal/Network/DNS/Decode/Parsers.hs
```diff
@@ -281,6 +282,13 @@ getRData NSEC3PARAM _ = RD_NSEC3PARAM <$> decodeHashAlg
     decodeIterations = get16
     decodeSalt       = getInt8 >>= getNByteString
 --
+getRData CAA len = do
+    dend <- rdataEnd len
+    flags <- get8
```
vdukhovni, line 287: Perhaps we should avoid the shadowing warning here???

PR #165, opened 2021-12-20
UDP ReceiveFrom, sendTo with SockAddr

Montycarlo:
The existing `Network.DNS.IO` functions `receive` and `send` do not facilitate normative UDP use; the client address is not returned by `receive` and thus `send` has no destination.

This PR adds the following two functions so that consumers of the library can respond to UDP messages using this library.


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
@vdukhovni I think this is reasonable additions. What do you think?

Network/DNS/IO.hs
```diff
@@ -36,19 +38,30 @@ import Network.DNS.Types.Internal
 
 ----------------------------------------------------------------
 
+-- | Receive and decode a single 'DNSMessage' from a UDP 'Socket', throwing away
```
vdukhovni, line 41: The original `receive` function is small enough that maintaining two copies is entirely reasonable.  The simpler function avoids unnecessary `sockaddr` allocation when the peer address is not needed.

Instead of reimplementing `receive` as a **wrapper** around `receiveFrom`, please just copy the `receive` function and apply the changes to that copy.  Though it should be safe to call `recvfrom` on a connected (UDP) socket, it is a common usage pattern, and I am also concerned about possible implementation bugs in the kernel (low probability, but best to avoid uncommon code paths).

stack.yaml
```diff
@@ -17,7 +17,7 @@
 #
 # resolver: ./custom-snapshot.yaml
 # resolver: https://example.com/snapshots/2018-01-01.yaml
-resolver: lts-12.24
+resolver: lts-18.19
```
vdukhovni, line 20: Did you need to change the resolver version?  Wha is the *oldest* LTS we want support?  Perhaps still the LTS for GHC 8.0?  That would actually be 9.21!   WIth 12.24 being GHC 8.4.4, I think we should let it stand at that or older for now???

vdukhovni (approved):
Code changes look good to me.  I don't have strong view about the LTS version, whatever Kazu san says...

stack.yaml
Montycarlo: Reverted change - originally I was receiving a build error but after cleaning `~/.stack` and `stack clean --full` and reverting version I no longer receive the error.

Original error was along the lines of hugs is no longer supported.

Montycarlo:
Requires approval for workflow to run.

stack.yaml
kazu-yamamoto: I don't have any opinions about the LTS version.

kazu-yamamoto:
Workflow is approved.

kazu-yamamoto:
Merged. Thanks.

If you would like to have a new release, I will.

PR #166, opened 2022-06-20
Introduce a flag to avoid service resolution

andreabedini:
Hi Kazu,

Haskell applications distributed as a statically linked executable typically use musl libc. This library hardcodes the `services` file location to `/etc/services` while some Linux distributions have started using GNU libc machinery (`/etc/nsswitch.conf`) to migrate `services` to `/usr/etc/services`.

While acknowledging that this is not because of a bug in the `dns` library,  using the `dns` library in the situation above would fail. 

Indeed, when resolving the address of the dns server we want to use we call `getAddrInfo` with the serivce name `domain`. This triggers a service resolution procedure provided by libc, which hits the different expectations between musl libc and gnu libc.

For consumers who need to fix this behaviour, it is useful to have a compile time flag that avoid this service resolution entirely and instead hardcodes the standard port 53 for DNS queries.

An example of this is cardano-node, which doesn't work at all on openSUSE, failing to contact the DNS server.:
```
[fino:cardano.node.DnsResolver:Error:233] [2022-06-15 06:08:50.66 UTC] Domain: "relays-new.cardano-mainnet.iohk.io" lookup exception Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG,AI_NUMERICHOST,AI_PASSIVE], addrFamily = AF_UNSPEC, addrSocketType = Datagram, addrProtocol = 0, addrAddress = 0.0.0.0:0, addrCanonName = Nothing}, host name: Just "10.43.10.1", service name: Just "domain"): does not exist (Unrecognized service)
```

I hope you can consider this patch. I am keen to hear what you think.

### DIFF
--- a/Network/DNS/Resolver.hs
+++ b/Network/DNS/Resolver.hs
@@ -67,12 +67,12 @@ makeResolvSeed conf = ResolvSeed conf <$> findAddresses
 
 makeAddrInfo :: HostName -> Maybe PortNumber -> IO AddrInfo
 makeAddrInfo addr mport = do
-    let flgs = [AI_ADDRCONFIG, AI_NUMERICHOST, AI_PASSIVE]
-        hints = defaultHints {
-            addrFlags = if isJust mport then AI_NUMERICSERV : flgs else flgs
+    let hints = defaultHints {
+            addrFlags = [AI_ADDRCONFIG, AI_NUMERICHOST, AI_NUMERICSERV, AI_PASSIVE]
           , addrSocketType = Datagram
           }
-        serv = maybe "domain" show mport
+        -- 53 is the standard port number for domain name servers as assigned by IANA
+        serv = maybe "53" show mport
     head <$> getAddrInfo (Just hints) (Just addr) (Just serv)
 
 ----------------------------------------------------------------


kazu-yamamoto (changes requested):
I'm sorry but I dislike this flag approach. What about adding a field to customize the service name or port in `ResolveConf`?

andreabedini:
@kazu-yamamoto here is a new design where I add a `ServiceOrPort` option to `ResolvConf`. I am not too found of the names I used so I am open to suggestions.

kazu-yamamoto (approved):
LGTM
@vdukhovni what about you?

vdukhovni:
> LGTM @vdukhovni what about you?

My take is that:

* The problem is entirely a base platform screwup.  Wherever the OS distribution puts the `services` file, that's where getAddrInfo **MUST** look for it.  Platforms where this is not the case are broken and unsupported.
* Haskell DNS should have non-trivial code to work around broken platforms.
* That said, I think the simplest solution is to just hardcode the numeric string "53" instead of "domain", and set the `AI_NUMERICSERV` flag. If some had a system where `domain` was some other port, that was a hack we need not support.

There's no need to rely on /etc/services to tell us that "domain" is shall be "53".  If at some point we implement support for DoT or DoH, then we can revisit how those service endpoints are resolved.

vdukhovni:
If there's a compelling use case for a non-default upstream resolver port or service name, then we can just test whether the input is numeric and use `AI_NUMERICSERV` accordingly.  If someone has, e.g.:
```
12345 54321/tcp
```
in their `/etc/services` file, that's too bad...

vdukhovni (changes requested):
So, bottom line, my take is that for now the best change is to just replace the literal `"domain"` with `"53"`, and hardcode `AI_NUMERICSERV` in the flags. And that's all.

kazu-yamamoto:
I have no objection to @vdukhovni's opinion.
@andreabedini What about you?

andreabedini:
I was originally aiming for the smallest change requiring the smallest level of decision making :) but I totally agree with @vdukhovni analysis. 

>  The problem is entirely a base platform screwup. Wherever the OS distribution puts the services file, that's where getAddrInfo MUST look for it. Platforms where this is not the case are broken and unsupported.
>  Haskell DNS should not have non-trivial code to work around broken platforms.

100% agree. In my particular case musl libc doens't work well on openSUSE. Full stop, it's their problem not ours. Done.

> That said, I think the simplest solution is to just hardcode the numeric string "53" instead of "domain", and set the AI_NUMERICSERV flag. If someone had a system where domain was some other port, that was a hack we need not support.

Even more, that would just be for the default port. Anybody can configure a resolver available on a different port if needed.

I'll update the PR. Thank you all.

vdukhovni:
Please rebase the PR.  Unfortunately the test suite relies on external DNS data that changes over time. :-)
I just updated the test that was failing because of upstream changes...

vdukhovni (approved):
CI has passed. Looks good.

kazu-yamamoto:

angerman:
🎉 

NB: I believe this is also what used to break dns resolution on android; but I'll have to confirm.

PR #167, opened 2022-06-22
add relaxed lookup-raw interface

khibino:
Add a new lookupRaw variant.
For example, the current time is not to be retrieved in the recv action, but passed from the outside.


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:

PR #170, opened 2022-09-06
Extensible RR

kazu-yamamoto:
I changed my mind. I tried to make RR extensible first. The key typeclass is `ResourceData `:

```
class (Typeable a, Eq a, Show a) => ResourceData a where
    encodeResourceData :: a -> SPut
    decodeResourceData :: proxy a -> Int -> SGet a
    copyResourceData   :: a -> a

data RData = forall a . (Typeable a, Eq a, Show a, ResourceData a) => RData a
```

I hope that @vdukhovni  likes this.

Currently, the parser calls `decodeResourceData` for known RRs only. We should discuss how to implement the table of `decodeResourceData` for unknown RRs.

Also, I would like to discuss whether or not we should make `OData` extensible in the same way.


vdukhovni:
This is analogous to design decisions I made in my dormant prototype, which makes further departures from the existing `dns` library, but I am rather concern that this may be too big a change for this library.

Any user of the existing API who goes beyond the high-level interface (`lookupMX`, `lookupA`, ...) immediately runs into major compatibility issues. This fundamentally changes the user-visible datatypes. Is it 5.0 of the existing library?  Or a new library, with the old library relegated to bugfix maintenance?

Perhaps, given PVP, and if enough users place version bounds in their cabal files, the change is a tolerable one-time barrier, but it is a big deal.  If you're willing to change this much, perhaps more should be changed as well, so that we don't keep introducing major changes repeatedly...

kazu-yamamoto:
OK. A new library would be better.
In the new library, `Domain` can be `ShortByteString` to avoid fragmentation.

kazu-yamamoto:
The following page shows that many packages does NOT specify the upper bound.

https://packdeps.haskellers.com/reverse/dns

kazu-yamamoto:
@vdukhovni FYI: I'm implementing new DNS libraries: https://github.com/kazu-yamamoto/dnsext

PR #172, opened 2023-06-30
Use showTXT to present the value of a CAA record.

vdukhovni:
The presentation form of the value for a CAA RR is the same as with DNS character strings (just not limited to 255 bytes).  Therefore, we need to use `showTXT` not `show` to present the value.


kazu-yamamoto (approved):
LGTM

kazu-yamamoto:
Merged.

